PR #26, opened 2015-05-15
Python 3 + New display options

rgerkin:


niklasf:
Hi @rgerkin, thanks for your pull request!

* I think it would be best to keep Python 3 compability fixes and new features seperated.
* Python 3 is generally already supported.

I'll provide some more specific feedback as comments to the commits.

Niklas

chess/__init__.py
```diff
@@ -26,12 +26,17 @@
 import re
 
 
-COLORS = [ WHITE, BLACK ] = range(2)
+COLORS = [ WHITE, BLACK ] = list(range(2))
```
niklasf: In general adding list(...) around ranges or views is not required. We're just iterating over them anyway. This should also apply to all other cases.

chess/uci.py
```diff
@@ -16,8 +16,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 
-from __future__ import print_function
-from __future__ import unicode_literals
```
niklasf: This is a valid cleanup.

chess/uci.py
```diff
@@ -29,7 +29,7 @@
 try:
     import queue
 except ImportError:
-    import Queue as queue
+    import queue as queue
```
niklasf: The uppercase `Queue` is intended as a fallback for Python 2.

chess/polyglot.py
```diff
@@ -158,7 +158,7 @@ def next_raw(self):
         except struct.error:
             raise StopIteration()
 
-    def next(self):
```
niklasf: I think this rename is correct. We should keep `next` as an alias for Python 2.

chess/__init__.py
```diff
@@ -2418,6 +2426,60 @@ def __str__(self):
 
         return "".join(builder)
 
+    def __unicode__(self):
+        string = '   a  b  c  d e  f  g  h  \r\n'
+        for rank in range(8,0,-1):
+            string += '  ' + '-'*22 + '\r\n'
+            string += '%d ' % rank
+            for file_ in 'A B C D E F G H'.split():
+                square = eval('%s%d' % (file_,rank))
```
niklasf: Let's not use `eval`.

niklasf:
Maybe a good next stop would be getting in the new `__unicode__()`.

rgerkin:
This pull request is now obsolete.

PR #30, opened 2015-05-19
Added html board

rgerkin:
You can display the board in an Ipython notebook with:  
```from IPython import display```

If you have a game running, you can continuously display the moves in one output cell with:                  
```display.clear_output(wait=True)
display.display(display.HTML(board.__html__()))
```

### DIFF
--- a/chess/__init__.py
+++ b/chess/__init__.py
@@ -2465,6 +2465,48 @@ class Board(object):
             string += '   a b c d e f g h'
         return string
 
+    def __html__(self):
+        """
+        Returns a html-version of the board.
+        Can be displayed in e.g. an IPython notebook using
+        IPython.display.HTML(board.__html__())
+        """
+        tr = '<tr style="vertical-align:bottom;">'
+        string = '<table style="text-align:center;\
+                                border-spacing:0pt;\
+                                font-family:\'Arial Unicode MS\';\
+                                border-collapse:collapse;\
+                                border-color:black;\
+                                border-style:solid;\
+                                border-width:0pt 0pt 0pt 0pt">'    
+        for rank in range(8,0,-1):
+            string += tr
+            string += '<td style="vertical-align:middle;\
+                                  width:12pt">%d</td>' % rank
+            for i,file_ in enumerate('a b c d e f g h'.split()):
+                square = SQUARE_NAMES.index('%s%d' % (file_,rank))
+                piece = self.piece_at(square)
+                char = piece.unicode_symbol() if piece else ''
+                if (i+rank) % 2 == 0:
+                    string += '<td style="width:28pt;\
+                                          height:28pt;\
+                                          border-collapse:collapse;\
+                                          border-color:black;\
+                                          border-style:solid;\
+                                          border-width:0pt 0pt 0pt 0pt">\
+                                          <span style="font-size:250%%;">\
+                                          %s</span></td>' % char
+                else:
+                    string += '<td style="background:silver;">\
+                               <span style="font-size:250%%;">\
+                               %s</span></td>' % char
+            string += '</tr>'
+        string += '<tr><td></td>'
+        for file_ in 'a b c d e f g h'.split():
+            string += '<td style="text-align:center">%s</td>' % file_
+        string += '</tr></table>'
+        return string
+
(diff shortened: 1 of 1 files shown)


niklasf:
I am on the fence about this once. Is having an `__html__` magic method (with double underscores) a common thing (like `__str__` or `__unicode__`)? Maybe an IPython convention?

rgerkin:
Are you on the fence about the value of the method, or just the naming convention?  I'm happy to have it changed to just html().  I find it useful to watch the games while I interactively code in the IPython notebook, without having to connect to some other GUI frontend.

niklasf:
Both.

First: Clearly if `__html__` is not a widely used convention, we shouldn't mark it as a magic method. So `html` would be fine.

Then: Something about the CSS and HTML seams so arbitrary to me.

I guess I don't fully appreciate the usefulness because I never used IPython notebooks. Then again having such an `html` method shouldn't hurt. Do you know some other libraries that provide a similar HTML feature to look at for reference?

rgerkin:
It's used somewhat frequently in web development libraries, e.g. [here](https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#safe=off&q=python+%22__html__%22).  But I'll change it to .html() if that's your vote.  

For the arbitrariness of the CSS and HTML, an alternative implementation would be to have a static template (a board.html file in the project) with tags like {{A1}, {{A2}}, and so on, and then the html method would use a template engine (jinja or Mako, or simply string.Template) to substitute the corresponding pieces (their unicode representations) or lack thereof.  This keeps content and style separate.  

As for the utility of it, it goes beyond an IPython notebook.  Any web interface using python-chess as part of the backend will want to have a board rendered in HTML.

niklasf:
Oh, so `__html__` is a thing. I didn't come up with anything useful when searching without the quotes. Let's get the simplest version in and maybe improve on it, later.

chess/__init__.py
```diff
@@ -2465,6 +2465,48 @@ def __unicode__(self, reverse_color=False, borders=False):
...
+                                border-collapse:collapse; \
+                                border-color: black; \
+                                border-style: solid; \
+                                border-width: 0pt 0pt 0pt 0pt">'    
+        for rank in range(8,0,-1):
+            string += tr
+            string += '<td style="vertical-align:middle; \
+                                  width:12pt">%d</td>' % rank
+            for i,file_ in enumerate('A B C D E F G H'.split()):
+                square = eval('%s%d' % (file_,rank))
```
niklasf: Let's avoid `eval`.

chess/__init__.py
```diff
@@ -2465,6 +2465,48 @@ def __unicode__(self, reverse_color=False, borders=False):
...
+        IPython.display.HTML(board.__html__())
+        """
+        tr = '<tr style="vertical-align:bottom;">'
+        string = '<table style="text-align:center; \
+                                border-spacing:0pt; \
+                                font-family:\'Arial Unicode MS\'; \
+                                border-collapse:collapse; \
+                                border-color: black; \
+                                border-style: solid; \
+                                border-width: 0pt 0pt 0pt 0pt">'
```
niklasf: Trailing spaces.

chess/__init__.py
```diff
@@ -2465,6 +2465,48 @@ def __unicode__(self, reverse_color=False, borders=False):
...
+        """
+        Returns a html-version of the board.
+        Can be displayed in e.g. an IPython notebook using
+        IPython.display.HTML(board.__html__())
+        """
+        tr = '<tr style="vertical-align:bottom;">'
+        string = '<table style="text-align:center; \
+                                border-spacing:0pt; \
+                                font-family:\'Arial Unicode MS\'; \
+                                border-collapse:collapse; \
```
niklasf: All other CSS properties have a space after the ':'

chess/__init__.py
rgerkin: Oops, I fixed this before but it reverted somehow... I think I branched from before the fix.  I'll take care of it.

niklasf:
Thank you! Committed and pushed. I took the liberty to remove the remaining trailing spaces.

PR #43, opened 2015-08-04
Add opposite color in Piece object

marcwebbie:


marcwebbie:
I agree with you @niklasf That was my feeling when I've seen that I should be "xoring", `WHITE` or `not WHITE` looks a lot more expressive.

niklasf:
The next release will have boolean constants for colors: #44.

PR #51, opened 2015-09-10
Enable Python 2.6 compatibility.

vdbergh:
In case the Python version is < 2.7 this patch enables the use 

backport_collections
unittest2

Both are available on pip.

On Python 2.6 mmap throws a different type of error. With this patch this error is correctly caught.

With these changes the test suite passes without errors (some tests are skipped but this is also true on my Python 2.7 system).

### DIFF
--- a/chess/polyglot.py
+++ b/chess/polyglot.py
@@ -72,10 +72,9 @@ class MemoryMappedReader(object):
 
         try:
             self.mmap = mmap.mmap(self.fd, 0, access=mmap.ACCESS_READ)
-        except Exception as e:
-            if type(e)==ValueError or (type(e)==mmap.error and e.errno==errno.EINVAL):
+        except (ValueError,mmap.error):
             # Can not memory map empty opening books.
-                self.mmap = None
+            self.mmap = None
 
     def __enter__(self):
         return self
--- a/chess/polyglot.py
+++ b/chess/polyglot.py
@@ -17,12 +17,19 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 import chess
-import collections
 import struct
 import os
 import mmap
 import random
 import itertools
+import chess
+import errno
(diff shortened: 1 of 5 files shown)


niklasf:
Good point finding all the points, especially the one with mmap!

Can we please use

```
try:
    import foo
except ImportError:
    import foo_backport
```

instead of looking directly for versions? Use `from ... import` whenever module level granularity is not enough.

chess/polyglot.py
```diff
@@ -65,9 +72,10 @@ def __init__(self, filename):
 
         try:
             self.mmap = mmap.mmap(self.fd, 0, access=mmap.ACCESS_READ)
-        except ValueError:
+        except Exception as e:
```
niklasf: Let's just catch both like `except (ValueError, mmap.error):` and not look closer at `e.errno`, if we can't handle it differently anyway.

vdbergh:
```
try:
    import foo
except ImportError:
    import foo_backport
```
I could try to cook up something like that but let me explain first. The problem is that 
```
    import collections
    import unittest
```
will not raise ImportErrors on Python 2.6, but the resulting modules will be missing methods.

Mixing methods ourselves from the official and the backported modules could actually be dangerous.

The authors of both backport_collections and unittest2 claim they fall back to collections and unittest 
for methods that are available. So these backported modules should be safe to use.

vdbergh:
Maybe you prefer something like this during initialization 
````
use_backport_collections=False
try:
    from collections import Counter as tmp
# maybe more import tests...
except ImportError:
    use_backport_collections=True
tmp=None
````
and then during import
````
if use_backport_collections:
    import backport_collections
else:
   import collections
````
To me it seems more code though.

niklasf:
I see, thanks! I guess your original code is best, then. Maybe I'll still change it after some careful testing.

PR #52, opened 2015-09-19
start work on native gaviota

jnavila:
This PR's purpose is only for following the on-going port to python of Gaviota. Do not base work on it as it will be force-pushed.

### DIFF
--- a/chess/gaviota_native.py
+++ b/chess/gaviota_native.py
@@ -1,4 +1,5 @@
 import os
+import struct
 from enum import Enum
 from lzma import LZMADecompressor
 import chess
@@ -1184,28 +1185,21 @@ def egtb_loadindexes(currentFilename, currentStream):
     idx = 0
 
     # Get Reserved bytes, blocksize, offset 
-    #currentStream.Seek(0, SeekOrigin.Begin)
     currentStream.seek(0)
-    dummy = fread32(currentStream)
-    dummy = fread32(currentStream)
-    blocksize = fread32(currentStream)
-    dummy = fread32(currentStream)
-    tailblocksize1 = fread32(currentStream)
-    dummy = fread32(currentStream)
-    tailblocksize2 = fread32(currentStream)
-    dummy = fread32(currentStream)
-    offset = fread32(currentStream)
-    dummy = fread32(currentStream)
+    HeaderStruct=struct.Struct("<10I")
+    Header = HeaderStruct.unpack(currentStream.read(HeaderStruct.size))
+
+    #blocksize = Header[2]
+    #tailblocksize1 = Header[4]
+    #tailblocksize2 = Header[6]
+    offset = Header[8]
 
     blocks = int((offset - 40) / 4) - 1
     n_idx = blocks + 1
 
-    p = [-1]*n_idx
-
-    # Input of Indexes 
-    for i in range(n_idx):  
-        idx = fread32(currentStream)
-        p[i] = idx # reads a 32 bit int, and converts it to index_t 
+    # Input of Indexes
+    indexStruct=struct.Struct("<" + "I"*n_idx)
+    p=indexStruct.unpack(currentStream.read(indexStruct.size))
 
     if (not currentFilename in Zipinfo):
         Zipinfo[currentFilename] = ZIPINFO()
--- a/chess/gaviota_native.py
+++ b/chess/gaviota_native.py
@@ -1,4 +1,22 @@
+
 import os
+# -*- coding: utf-8 -*-
+#
+# This file is part of the python-chess library.
+# Copyright (C) 2015 Jean-Noël AVILA<jn.avila@free.fr>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
 import struct
 from enum import Enum
 from lzma import LZMADecompressor
@@ -20,8 +38,8 @@ class currentConf(object):
     def __init__(self):
         self.whitePieceTypes = None
         self.blackPieceTypes = None
-        self.whitePieceSquares = [] # new List<int>();
-        self.blackPieceSquares = [] # new List<int>();
+        self.whitePieceSquares = [] 
+        self.blackPieceSquares = []
         self.Reversed = False
 
 NOINDEX = -1
@@ -66,14 +84,11 @@ INFOMASK = 7
 currentPctoi = None
 
 class endgamekey:
-
     def __init__(self, a_maxindex, a_slice_n, a_pctoi):
         self.maxindex = a_maxindex
         self.slice_n = a_slice_n
         self.pctoi = a_pctoi
 
-def EGTB():
-    pass
 
 WE_FLAG = 1
 NS_FLAG = 2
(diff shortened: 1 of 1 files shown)


jnavila:
Right now the code barely passes syntax check, but is utterly broken.

chess/gaviota_native.py
```diff
@@ -0,0 +1,2436 @@
...
+    dtm = 0
+
+entries_per_block = 16 * 1024;
+currentStream = None
+currentFilename = ""
+validTables = [] # new List<string>(15);
+a8toa1 = {} #new Dictionary<int, int>();
+
+
+A1 = 0
```
niklasf: This happens to be the same square indexing scheme as the already defined `chess.A1`, ..., `chess.H8`.

chess/gaviota_native.py
```diff
@@ -0,0 +1,2436 @@
...
+class TableBlock:
+
+    def __init__(self, currentFilename, stm, o):
+        self.key = currentFilename
+        self.side = stm
+        self.offset = o
+        self.pcache = []
+
+    # TODO: not sure this is really useful...
+    def __hash__(self):
```
niklasf: As long as the `blockCache` dictionary does not have TableBlocks as *keys* this is not required.

niklasf:
Good job! I'll have a closer look, later. Feel free to ignore the comments at this stage.

chess/gaviota_native.py
jnavila: At first sight, there are some chunks of code that are not useful.

Some global variables are shared deeply into the call graph.

For the initialization, I expect to just initialize the tables at the time they are declared; this means that the import of the module will add a initialization overhead. Is it affordable?

For the probing, I have two options:
 * functional way: the variables are reintegrated in the functions' parameters. This makes the use of the variables clearer, but the functions' prototypes may become long. The code will derive from the original.
 * Object oriented way: the probe functions are inserted into a class and the needed global variables become attributes. This may be the simplest way.

The second option seems to be more in the spirit of the existing code.

Finding where to apply the changes is quite simple: pyflakes warns when a variable will not survive the context.

niklasf: Whatever way of initialization is easiest is fine. I can always refactor it, later.

If you need the same "global" state often, I'd definitely prefer the object oriented approach.

jnavila:
Hello @niklasf ! Some good news!

With this new release, I was able to probe some samples such as:

```python
import chess.gaviota_native as g
import chess
gv=g.PythonTableBase("chess/gaviota_tables/")

b =chess.Board("8/8/8/8/8/8/8/K2kr3 w - - 0 1") # -10
a = gv.probe_dtm(b)
assert(a.dtm==-10)

b =chess.Board("8/8/8/8/4k3/8/6R1/7K w - - 0 1") # 31
a = gv.probe_dtm(b)
assert(a.dtm==31)

b =chess.Board("8/2K5/4B3/3N4/8/8/4k3/8 b - - 0 1") #White to mate , 54
a = gv.probe_dtm(b)
assert(a.dtm==-54)

b =chess.Board("8/2k5/4b3/3n4/8/8/4K3/8 b - - 0 1") # Black to mate 51
a = gv.probe_dtm(b)
assert(a.dtm==51)
```


I get the same results as the Csharp. But there is a difference in both implementations with syzygy's result taken from your readme page (8/2K5/4B3/3N4/8/8/4k3/8 b - - 0 1). Even worse,when tested with "7P/2k5/4b3/3n4/8/8/4K3/8 b - - 0 1", the C# and the native_python fail miserably with an out of index exception. This basically boils down to saying that the C# was buggy in the first place. Can you check if the C version can resolve this one, and to which value?

Another front, the present implementation is python3 only for several reasons:

 * interpreting bytes as integers instead of chars. This should disappear with the switch to struct.
 * using lzma directly from the standard library starting from python 3.2. For previous versions, an additional library is needed.

The code is still not ready for general consumption.

jnavila:
OK, forget about the index out of bounds: the C version malloc's to the size announced in the file instead of allocating a fixed size for the index. Changing the python code, I get "draw".

niklasf:
Hi @jnavila. Excellent work!

I took some data from http://kirill-kryukov.com/chess/longest-checkmates/longest-checkmates.shtml to create a [test suite](https://github.com/niklasf/python-chess/commit/e739476cbf48be8254e60b5625d0a0f575c684f9#diff-d1642864be99e3461a5569d4a500fb22R1). You can merge it from my master branch. Most of the positions work, already.

One that doesn't:
```
k7/8/4K2p/8/8/8/8/n1n5 b - - got 55; line 44; dtm 55;
Traceback (most recent call last):
  File "test_gaviota.py", line 39, in <module>
    probe = tables.probe_dtm(board)
  File "/home/niklas/Projekte/python-chess/chess/gaviota_native.py", line 2156, in probe_dtm
    return self._Probe(side,  self.epsq)
  File "/home/niklas/Projekte/python-chess/chess/gaviota_native.py", line 2181, in _Probe
    is_dtm, dtm = self.egtb_get_dtm(side, epsq)
  File "/home/niklas/Projekte/python-chess/chess/gaviota_native.py", line 2291, in egtb_get_dtm
    ok, dtm  = self.tb_probe_(side, epsq)
  File "/home/niklas/Projekte/python-chess/chess/gaviota_native.py", line 2457, in tb_probe_
    dtm = t.pcache[remainder]
TypeError: list indices must be integers, not float
```

> The code is still not ready for general consumption.

Anything else (besides this and `probe_wdl`) before this is good to go? I won't hold you up to clean-ups.

jnavila:
Hi, @niklasf 

I fixed a couple things. Still, the code does not pass the tests you provided. I get a LZMA exception while reading "knpkp.gtb.cp4". The same as with the Csharp... I hadn't taken the time to compile the C version, but it seems it is needed to understand what's going on.

jnavila:
Now, all the test you provided are passing. Least to say, the C# porting has introduced a lot of bugs, by not copying correctly from the C original.

Some bugs may still be lurking around. Time to tidy up a little before handing it to you.

jnavila:
Hi @niklasf ,

I have a few open questions left:

 * The LRU cache has not been ported (the cache is never tidied). Is this a problem?
 * In the C# port, some table files entries are commented out. The tests did not choke on a missing table. Is there something more to do on this.
 * I did not check if the tests that you added make coveralls happy.

As for this PR, would you prefer that I squash and rebase it on top of `master` ?

niklasf:
* Yeah. The LRU cache should at least be in a state, where it would not slowly exhaust memory in a long running process.
* The entries that are commented out appear to be duplicate entries. I don't understand why the C# people did that, but it looks like they can be safely removed.
* Test coverage was pretty good, except for en passant positions. I added (6c3c72060e98) some extra endgames and indeed `1k1K4/8/8/pP6/8/8/8/8 w - a6` fails.

```
Traceback (most recent call last):
  File "test_gaviota.py", line 41, in <module>
    probe = tables.probe_dtm(board)
  File "/home/niklas/Projekte/python-chess/chess/gaviota_native.py", line 1829, in probe_dtm
    return self._Probe(side,  self.epsq)
  File "/home/niklas/Projekte/python-chess/chess/gaviota_native.py", line 1854, in _Probe
    is_dtm, dtm = self.egtb_get_dtm(side, epsq)
  File "/home/niklas/Projekte/python-chess/chess/gaviota_native.py", line 2011, in egtb_get_dtm
    ys, yp = removepiece(ys, yp, j)
TypeError: 'NoneType' object is not iterable
```

jnavila:
OK, now the en passant position passes. But, I am uneasy on the fix. The C# implementation was bugged too. The C original is really different in this area. So, I made the simplest modification that both seemed understandable and made the test pass.  The reverse logic in the test has bug that prevents from testing the other way around. Where can we find more of these?

For the LRU, I planned initially to use the lru_cache decorator on a well behaved function, but the code does not seem to allow this, so I'll just implement the usage age caching. Here again, I'm pretty sure the C# is bugged...

jnavila:
@niklasf OK. All done. If you want to test more, please do so.

niklasf:
Hi @jnavila, excellent. Merged! :+1: 

Thanks for this awesome contribution. Don't forget to [claim](https://www.bountysource.com/issues/26152993-pure-python-gaviota-tablebase-support) the bounty that was on this issue.

I'll add more tests and move stuff around as I go along.

PR #54, opened 2015-10-05
Add variation_san to create standard algebraic notation for a variation

amdw:
This pull request does for variations what board.san() does for individual moves.


niklasf:
Thank you, looks like a useful feature. Good job on the exhaustive test coverage! Merged.

amdw:
My pleasure - thank *you* very much for this nice library!

till314:
How about adding an option show_nags (defaulting to False) to variation_san which, if set, add the standard "!","?", etc to the output

PR #64, opened 2016-01-18
Fix case that breaks when loading some PGN's, since accessing game.board() raises ValueError.

balparda:
Fix case that breaks when loading some PGN's, since accessing game.board() raises ValueError.


niklasf:
Thanks! Merged.

PR #101, opened 2016-11-23
Add giveaway variant testsuite from Chess Variants Training

ddugovic:
Puzzles provided from [Chess Variants Training - Antichess](http://chessvariants.training/Puzzle/Antichess) with best moves generated by Stockfish

There's probably a cleaner way to code this test, but I don't see how.

### DIFF
--- a/examples/giveaway_bratko_kopec.py
+++ b/examples/giveaway_bratko_kopec.py
@@ -1,0 +1,233 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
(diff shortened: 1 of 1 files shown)


niklasf:
Thanks. Seems to me that there should be something more elegant than copying most of the script. I'll ponder that a little.

niklasf:
Ok ... so I added the testsuite in an extra file instead. Now it can be used like this:

```
python bratko_kopec.py -e ./stockfish-x86_64-bmi2 --variant giveaway giveaway-puzzles.epd
```

PR #117, opened 2017-02-11
SVG Tiny support

Self-Perfection:
Lots of applications support only SVG Tiny profile.
Most notable is Qt framework:
http://doc.qt.io/qt-5/svgrendering.html
> Qt supports the static features of SVG 1.2 Tiny.

And unfortunately SVG Tiny does not support <style> tag:
https://www.w3.org/TR/SVGMobile/#sec-styling
> SVGB allows optional support of CSS Mobile Profile 1.0.
> SVGT does not support styling with CSS.

Therefore before this change most apps show all squares black.

This change adds fallback properties with colour for board squares to
allow tolerable rendering with engines that do not support <style>.

If app supports <style> tag, it overrides directly specified properties
so adding fill color to squares act only as fallback.

---
Actually before this patch only webbrowsers and Inkscape displayed SVG board generated by python-chess rather properly. Everything else showed just black board. For instance:
jupyter qtconsole_2


niklasf:
Thank you. I'll merge this now and see if I can also remove the default CSS.

Issue #116, opened 2017-02-10
SVG questions & feature requests


LocutusOfPenguin:
Hi,
I have a bunch of questions & feature requests:

- is the SVG support more or less guaranteed to continue in future releases? (background: i can remember its said "experimental" like its now with variants).
- can the svg board be made responsive? If yes, how? I see you need to give Width,Height coordinates at the startup inside (but i prefer it scales on a surrounding div for example)
- I would like to draw arrows (like its on lichess). How to do this? Or can you provide an api for this?

Jürgen

niklasf:
Hi Jürgen, I think SVG support is quite useful (especially in combination with IPython), so it's here to stay. As for the feature requests, both are possible, but it might be a while before I get to implement them. Niklas

niklasf:
Thanks to @rheber we now have support for arrows.

LocutusOfPenguin:
that's really great. Thanks!

niklasf:
`size` now defaults to `None` in `chess.svg.board(..., size=None)` and will produce scalable SVGs with a `viewBox`.

That should conclude this issue.

PR #122, opened 2017-03-12
Basic SVG arrows

rheber:
#116 
I imagine there's a fair bit that can be improved but the basic functionality is there.

### DIFF
--- a/chess/svg.py
+++ b/chess/svg.py
@@ -22,6 +22,10 @@
 
 import chess
 
+try:
+    import backport_collections as collections
+except ImportError:
+    import collections
 
 PIECES = {
     "b": """<g id="black-bishop" class="black bishop" fill="none" fill-rule="evenodd" stroke="#000" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"><path d="M9 36c3.39-.97 10.11.43 13.5-2 3.39 2.43 10.11 1.03 13.5 2 0 0 1.65.54 3 2-.68.97-1.65.99-3 .5-3.39-.97-10.11.46-13.5-1-3.39 1.46-10.11.03-13.5 1-1.354.49-2.323.47-3-.5 1.354-1.94 3-2 3-2zm6-4c2.5 2.5 12.5 2.5 15 0 .5-1.5 0-2 0-2 0-2.5-2.5-4-2.5-4 5.5-1.5 6-11.5-5-15.5-11 4-10.5 14-5 15.5 0 0-2.5 1.5-2.5 4 0 0-.5.5 0 2zM25 8a2.5 2.5 0 1 1-5 0 2.5 2.5 0 1 1 5 0z" fill="#000" stroke-linecap="butt"/><path d="M17.5 26h10M15 30h15m-7.5-14.5v5M20 18h5" stroke="#fff" stroke-linejoin="miter"/></g>""",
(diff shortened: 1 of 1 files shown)


chess/svg.py
```diff
@@ -58,6 +59,13 @@
 }}
 """.format(**DEFAULT_COLORS)
 
+class Arrow():
+    """
+    Details of an arrow to be drawn.
+    """
+    def __init__(self, head, tail):
```
niklasf: This is bikeshedding, but the order `tail -> head` like `Move(from_square, to_square)` seems more natural to me.

chess/svg.py
```diff
@@ -80,7 +88,7 @@ def piece(piece):
     return _svg(PIECES[piece.symbol()], 45, 45)
 
 
-def board(board=None, squares=None, flipped=False, coordinates=True, lastmove=None, check=None, size=400, style=None, pre="", post=""):
+def board(board=None, squares=None, flipped=False, coordinates=True, lastmove=None, check=None, size=400, style=None, pre="", post="", arrows=[]):
```
niklasf: Lets have this after `check` and before `size`. (Technically that would break backwards compability, but these are basically keyword-only arguments).

chess/svg.py
```diff
@@ -58,6 +59,13 @@
 }}
 """.format(**DEFAULT_COLORS)
 
+class Arrow():
```
niklasf: How about making this a `namedtuple` like so: https://github.com/niklasf/python-chess/blob/master/chess/uci.py#L69? And then allowing the more succint `board(..., arrows=[(E2, E4)])` would be nice.

chess/svg.py
```diff
@@ -162,5 +173,17 @@ def board(board=None, squares=None, flipped=False, coordinates=True, lastmove=No
...
+        head_file = chess.file_index(arrow.head)
+        head_rank = chess.rank_index(arrow.head)
+        tail_file = chess.file_index(arrow.tail)
+        tail_rank = chess.rank_index(arrow.tail)
+
+        xhead = (head_file + 1 if not flipped else 8 - head_file) * square_size
+        yhead = (8 - head_rank if not flipped else head_rank + 1) * square_size
+        xtail = (tail_file + 1 if not flipped else 8 - tail_file) * square_size
+        ytail = (8 - tail_rank if not flipped else tail_rank + 1) * square_size
+        builder.append("""<line x1="%f" y1="%f" x2="%f" y2="%f" stroke="#888" stroke-width="4" marker-end="url(#arrowhead)"/>""" % (xtail, ytail, xhead, yhead))
```
niklasf: image

A couple points here:

* The arrow seems slightly off center.
* The arrow head is a bit flat.
* The arrow width does not scale with the board size.

In terms of what do aim for, I very much like the much wider (but transparent) arrows of lichess:

image

rheber:
I've addressed most of your critiques but regretfully I've been unable to get the arrow centering quite right.

niklasf:
Thank you very much. It took me a while to figure out, but the reason the arrows were slightly off was the margin with the coordinates around the board.

I pushed a fix to your branch.

I did however break your handling of the `head = target` case. Maybe we should draw a circle on the square in that case? Or just skip that arrow ...

niklasf:
I think the result is quite nice now:

demo svg

Merged. Thanks again :)

PR #129, opened 2017-05-24
Xboard

niklasf:
Hi @Mk-Chan, and thanks in advance. I've created this PR as a reminder for myself to review your changes.


chess/__init__.py
```diff
@@ -626,14 +626,15 @@ def __init__(self, from_square, to_square, promotion=None, drop=None):
         self.promotion = promotion
         self.drop = drop
 
-    def uci(self):
+    def pacn(self):
         """
-        Gets an UCI string for the move.
+        Gets an PACN string for the move.
```
niklasf: What does `PACN` stand for? I couldn't find the term in the XBoard protocol spec (or anywhere else).

chess/xboard.py
```diff
@@ -0,0 +1,423 @@
...
+            with self.semaphore:
+                self.send_line("xboard")
+                self.ping(randint(0, 100))
+
+                if self.terminated.is_set():
+                    raise EngineTerminatedException()
+
+        return self._queue_command(command, async_callback)
+
+    def new(self, board, async_callback=None):
```
niklasf: `board` does not seem to be used. Is there another way to setup a specific position?

chess/__init__.py
Mk-Chan: That was a mistake on my part, PACN stands for Pure Algebraic Coordinate Notation
I got it from [The Chess Programming Wiki](https://chessprogramming.wikispaces.com/Algebraic+Chess+Notation#Pure%20coordinate%20notation)

chess/xboard.py
Mk-Chan: Well, the `new` command in xboard will reset the board to the starting position(Will implement it).
The other command is `setboard <fen>`

Mk-Chan:
I'll be using Crafty and my own engine, WyldChess(which does not have as many features of xboard yet :p) to test it. I wanted your view on if this sort of code transformation was fine. I've retained all the old functions to avoid any backwards compatibility issues. Also it seems like it's failing tests due to timeouts?(If i understand correctly). I'm not so sure why that is happening.

I'll begin testing with an engine to confirm the commands work correctly once I implement the `go` command

Mk-Chan:
The thing with xboard is that the engine itself keeps state so you don't have to copy `board` into it each time. Once you set it up with `setboard <fen>` or `new`, the engine will maintain it's own internal board so you can simply `undo` and `usermove <pacn-move>` to undo and play moves.

chess/__init__.py
cash: @niklasf do you have another suggestion for this? xboard and uci both use this coordinate notation for moves so calling it uci seems misleading...but there is no standard name for it either.

cash:
There is shared code between the two engine implementations. Should that be pulled out into a parent class?

Mk-Chan:
Yes, but also retain the chess.uci.Engine and the chess.xboard.Engine signatures so i guess sort of a overriding the class(not sure how to do it in python) and make the quit,kill,popen_engine functions member functions perhaps?

cash:
It could look like this:

```python
class Engine(object):
    def quit(self):
        # send quit message

class XBoardEngine(Engine):
    def go(self, ...):
        # unique move code for xboard

class UCIEngine(Engine):
    def go(self, ...):
        # unique move code for uci
```
I chose different class names for the xboard and uci engines in this example. I don't know if that is something @niklasf wants or not.

Mk-Chan:
Yea I suggest we wait for his input on that. Since that's just a refactor, I'll go ahead implementing the protocol as is for now.

chess/__init__.py
Mk-Chan: It is kind of odd calling it from_uci and from_xboard. I abstracted the implementation out so we could mark it deprecated or something while phasing it out(if we are actually going to remove it)

niklasf:
Hi, I think the best way forward is just to merge this and go from there. However I think it's best to just use `from_uci` and co. for now. Can you please revert those changes?

Mk-Chan:
Revert which one? iirc, I retained the from_uci and other similar function and also wrote from_xboard and the like

niklasf:
I mean the addition of from_pacn, pacn, and the conversion of uci to pacn.

Mk-Chan:
So just to confirm, remove all the pacn stuff and leave only the from_uci and from_xboard and the like(can just use self.from_uci(), etc in the xboard function definitions)?

niklasf:
Yeah. That should make the patch more self contained and we can consider the UCI -> PACN rename later (and independently).

Mk-Chan:
Ok, I did that, I guess next in line is filling in the test.py with some xboard engine tests. Merge first and go from there or should I write some tests?

niklasf:
Thanks :)

I definitely want to have this in the next release, so let's just merge.

PR #130, opened 2017-06-01
Update to the xboard branch

cash:
Updated the documentation based on the changes in the branch so that it passes the doc checks. Also fixed some bugs when testing it against an xboard engine.

@MK-Chan I can create a pull request against your branch if you want.

Sunfish might be a simple xboard implementation to use for integration tests: https://github.com/thomasahle/sunfish


Mk-Chan:
Should be fine to create a pull request against my branch, I was planning on using Crafty and WyldChess  already, but more engines don't hurt thanks!

PR #131, opened 2017-06-03
add kwargs to popen_engine()

cash:
I'm using a chess engine that requires it to be run in its directory to access its data files. This means I need to be able to pass cwd to subprocess.Popen(). This pull request adds support for that.


niklasf:
Thanks, look good.

PR #133, opened 2017-06-10
Opening classification

hauva69:
I digged up some of my very old code for opening classification.

The classification data comes from Andrew Tridgell's Lasker chess server and he considers that the data is GPL compatible. Unfortunaly he does not state the exact GPL version in https://github.com/ddugovic/capablanca/blob/master/README but I do believe in this case that if the data is GPL v2 compatible it's also GPL v3 compatible.

The original data has internationalization problems, for example Latvian master Alvis Vītoliņš is spelled both "Vitolins" and "`Vitoli\'n\'s`". I modified data to make "`R\'eti`" to "Réti" etc, but I'm quite sure that the data has many spelling problems. The data in data directory is the original, openingclassification.py has my editions.


ddugovic:
I wonder if positions which are equivalent with [this pawn structure encoder]9http://gist.github.com/ddugovic/e1381cfe60843f1aab15f880e6ae4eea) should have the same opening code & name.  For example 1. a3 a6 2. e4 e5 3. f4 exf4 would be a King's Gambit.

niklasf:
Thank you for the pull request. The significant part of this is isn't really specific to Python, so I think it should rather be a seperate dataset.

hauva69:
I did consider moving the data into separate files. If I make that change, are you interested in the pull request?

niklasf:
Imo it's most useful in a seperate repo that can be used from all kinds of programming languages.

niklasf:
Btw, here is another data set from various sources https://github.com/niklasf/eco/blob/master/eco.json. We use this one on https://lichess.org. I should see how yours compares and if any of the corrections are relevant ...

PR #134, opened 2017-06-18
replaces copied functions with internal functions in engine module

cash:
Two functions were duplicated in the xboard and uci modules. This pulls the functions into engine.py and puts a wrapper in its place. I couldn't just move the function into engine.py because of the default argument for engine_cls.


niklasf:
Looks good. Thanks :)

PR #139, opened 2017-07-02
Add Bratko-Kopec thread count parameter (rename movetime alias)

ddugovic:


examples/bratko_kopec/bratko_kopec.py
```diff
@@ -26,6 +26,7 @@ def test_epd(engine, epd, VariantBoard, movetime):
 
     engine.ucinewgame()
     engine.setoption({"UCI_Variant": VariantBoard.uci_variant})
+    engine.setoption({"Threads": threads})
```
niklasf: You can use a single `engine.setoption` here. Example: https://python-chess.readthedocs.io/en/latest/variant.html#uci

examples/bratko_kopec/bratko_kopec.py
```diff
@@ -63,7 +65,6 @@ def test_epd_with_fractional_scores(engine, epd, VariantBoard, movetime):
     score = 0.0
 
     print("%s:" % epd_string, end=" ")
-    sys.stdout.flush()
```
niklasf: Why did you remove this line?

ddugovic: I think I had originally added the `flush` call because I didn't know how to run python in the background with stdout unbuffered (`python -u`).  This code seems unnecessary when running in the foreground, since stdout is unbuffered.

niklasf: Ah, then no. These have to stay because stdout is line-buffered by default and we print progress as parts of a single line, here. Nicer would be `print(..., flush=True)`, but unfortunately that's not supported in Python 2.

niklasf:
Alright, merged. Pretty neat feature.

PR #140, opened 2017-07-04
XBoard support

Mk-Chan:
Usual improvements and bug fixes and finally wrote some tests.

### DIFF
--- a/test.py
+++ b/test.py
@@ -2100,7 +2100,7 @@ class CraftyTestCase(unittest.TestCase):
 
     def setUp(self):
         try:
-            self.engine = chess.xboard.popen_engine("/home/mkchan/git/crafty-chess/crafty")
+            self.engine = chess.xboard.popen_engine("crafty")
         except OSError:
             self.skipTest("need crafty")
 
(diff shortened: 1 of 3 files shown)


Mk-Chan:
The travis build for Python 2.6 failed because of a timeout....

Turns out the build of crafty I used does not respond to `ping 0` but works with just about any other number.

The protocol states `ping N: In this command, N is a decimal number`. So I don't think crafty is right here.

test.py
```diff
@@ -2096,6 +2096,49 @@ def test_variants(self):
...
+        except OSError:
+            self.skipTest("need crafty")
+
+        self.engine.xboard()
+        self.engine.send_line("log off")
+
+    def tearDown(self):
+        self.engine.quit()
+
+    def testSt(self):
```
niklasf: The other tests methods are currently named in `snake_case` (except setUp/tearDown, where there is no choice).

Mk-Chan: Alright, I'll do those and I guess I can add some obvious mate-in-x positions that crafty should not have a problem finding.

Mk-Chan:
Regarding the `ping <num>`, I contacted Bob and he said that it is as he and Tim intended for the Xboard spec. Apparently they had specified a 'positive integer' and it is implemented in Crafty to respond with `pong <num>` for any non-zero number. As such the change I had done to conform to crafty is (quite luckily) correct.

PR #141, opened 2017-07-07
Code improvements, tests and new func: playother()

Mk-Chan:
Interesting to note that occasionally the Travis build fails spuriously in one of the `script` statements(in travis.yml). For example [here](https://travis-ci.org/niklasf/python-chess/jobs/250055472#L1465)


niklasf:
Thanks.

PR #148, opened 2017-07-28
More commands

Mk-Chan:
I'm not exactly sure how to test `memory` and `cores`. Any suggestions?

### DIFF
--- a/chess/xboard.py
+++ b/chess/xboard.py
@@ -355,20 +355,41 @@ class Engine(object):
         """
         self.pondering(True)
 
-    def post(self, async_callback=None):
+    def set_post(self, flag, async_callback=None):
         """
-        Command used to tell the engine to output it's analysis
+        Command used to tell the engine whether to output it's analysis or not.
+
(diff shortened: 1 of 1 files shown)


Mk-Chan:
On the other hand, it is just an API, it's really up to the engine so is there even a need to test such functions?

chess/xboard.py
```diff
@@ -355,21 +355,42 @@ def hard(self, async_callback=None):
...
 
         return self._queue_command(command, async_callback)
 
+    def post(self, async_callback=None):
+        """
+        Command used to tell the engine to output it's analysis.
+
+        :return: Nothing
+        """
+        self.set_post(True)
```
niklasf: This should probably be `return self.set_post(True)`, so that it can return a `Future` depending on `async_callback`.

Mk-Chan: good catch, I think there's one more place i may have made the same mistake let me check

niklasf:
Thank you, merged!

PR #158, opened 2017-08-31
FeatureMap for Xboard

Mk-Chan:
This is still incomplete, but I wanted to make the PR mainly because of the first commit which modifies uci.py slightly. I still have to put in the check for features that call functions like `cores` to see if the feature is available with the engine or not. The option map parsing is also incomplete at the moment and I also need to include some tests into test.py.

### DIFF
--- a/chess/engine.py
+++ b/chess/engine.py
@@ -57,6 +57,11 @@ class EngineStateException(Exception):
     pass
 
 
+class Option(collections.namedtuple("Option", ["name", "type", "default", "min", "max", "var"])):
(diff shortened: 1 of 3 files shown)


Mk-Chan:
Added checks before feature API calls. Now need to complete OptionMap parsing

niklasf:
Looks good (and man, I had almost forgotten how much more complex XBoard is than UCI). I'll just go ahead and merge this as is, no need to block on the remaining TODOs.

PR #161, opened 2017-09-04
Added api for modifying engine options and some tests with wyld

Mk-Chan:
Had to add another engine for testing options since the crafty I'm using doesn't have any. This should complete the base feature/option API as long as I haven't missed anything

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,12 @@ before_install:
  - make unix-gcc
  - export PATH="`pwd`:${PATH}"
  - cd ..
+ - # WyldChess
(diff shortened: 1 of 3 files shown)


Mk-Chan:
Oh wow! It failed somehow! In one paticular build that too. Interestingly on my travis-ci it worked just fine [here](https://travis-ci.org/Mk-Chan/python-chess).
Could even be a problem in the engine because it seems to have stopped after `quit` but then I see that it has in fact said `ok` to the test. I'm at a bit of a loss here.

niklasf:
Thanks.

Looks like the test failure is not related to this patch anyway. So it's probably best to merge and figure it out later ...

PR #165, opened 2017-09-12
Added stop() command and removed buggy option testing

Mk-Chan:
It's probably not a good idea to use a time-based option because that's likely what's causing the bug. As such, I've removed it for now, will look for a different way/option to test that.


Mk-Chan:
[Here](https://travis-ci.org/Mk-Chan/python-chess/jobs/274640073) is the failed build log, it failed at exactly the same location for the same reason. Though I'm still not sure exactly what the reason is.

niklasf:
Ok, fair enough. Thanks :+1:

PR #172, opened 2017-10-21
Improve class Arrow in chess.svg

PedanticHacker:
A subclass of `collections.namedtuple` must implement the `__slots__`
attribute and, in this case, an empty tuple must be assigned to it.

### DIFF
--- a/chess/svg.py
+++ b/chess/svg.py
@@ -72,7 +72,7 @@ DEFAULT_STYLE = """\
 
 class Arrow(collections.namedtuple("Arrow", ["tail", "head"])):
     """Details of an arrow to be drawn."""
-    pass
+    __slots__ = ()
 
 
 def _svg(viewbox, size):


niklasf:
Thanks, I wasn't aware of this. Can you please explain? Is this required for correctness, is it an optimization?

PedanticHacker:
I read the Python documentation regarding collections.namedtuple. It is expressed there that when subclassing collections.namedtuple, the __slots__ = () is needed because "this helps keep memory requirements low by preventing the creation of instance dictionaries".

niklasf:
Cool. Maybe avoiding the subclassing is even better, since nothing is added anyway?

PedanticHacker:
I agree. Just remove this `Arrow` subclass object altogether and add `Arrow = collections.namedtuple("Arrow", ["tail", "head"])` somewhere in chess.svg module.

I was wondering why the field names `"tail"` and `"head"` are passed in as a list of strings. `Arrow = collections.namedtuple("Arrow", "tail, head")` does the trick as well. You can even pass them in as `"tail head"`, but I'd go for `"tail, head"` (my personal taste is that with the comma looks better).

niklasf:
Unfortunately this seems to not include the docstring (unless explicitly assigned to `__doc__`). So I'll just go with your first suggestion.

PR #175, opened 2017-10-22
Improve chess.svg.board docs

PedanticHacker:
Add an example of the exact code that the programmer would need to pass
in to generate an arrow on the chessboard.


niklasf:
Thanks.

PR #176, opened 2017-10-22
Another improvement of chess.svg.board docs

PedanticHacker:
These elaborations on the use of `chess.svg.board` function that I provide in this pull request are important for a user of python-chess to read. I wasted too much time because the documentation was not detailed.

### DIFF
--- a/chess/svg.py
+++ b/chess/svg.py
@@ -130,10 +130,13 @@ def board(board=None, squares=None, flipped=False, coordinates=True, lastmove=No
     :param lastmove: A :class:`chess.Move` to be highlighted.
     :param check: A square to be marked as check.
     :param arrows: A list of :class:`~chess.svg.Arrow` objects like
-        ``[chess.svg.Arrow(chess.E2, chess.E4)]``. Arrows from a square to the
-        same square are drawn as circles.
-    :param size: Limit the SVG size by setting width and height of the image.
-    :param style: CSS to use instead of the default stylesheet.
+        ``[chess.svg.Arrow(chess.E2, chess.E4)]`` or a list of tuples like
+        ``[(chess.E2, chess.E4)]``. An arrow from a square that is pointing to
+        the same square is drawn as a circle, like ``[(chess.E2, chess.E2)]``.
+    :param size: Adjust size of the SVG image by setting its width and
+        height, like ``(600, 600)``.
+    :param style: Add CSS markup to change the default style sheet of the SVG
+        image.
 
     >>> import chess
     >>> import chess.svg


chess/svg.py
```diff
@@ -130,10 +130,13 @@ def board(board=None, squares=None, flipped=False, coordinates=True, lastmove=No
     :param lastmove: A :class:`chess.Move` to be highlighted.
     :param check: A square to be marked as check.
     :param arrows: A list of :class:`~chess.svg.Arrow` objects like
-        ``[chess.svg.Arrow(chess.E2, chess.E4)]``. Arrows from a square to the
-        same square are drawn as circles.
```
niklasf: Good idea to add more concrete examples.

chess/svg.py
```diff
@@ -130,10 +130,13 @@ def board(board=None, squares=None, flipped=False, coordinates=True, lastmove=No
...
-    :param size: Limit the SVG size by setting width and height of the image.
-    :param style: CSS to use instead of the default stylesheet.
+        ``[chess.svg.Arrow(chess.E2, chess.E4)]`` or a list of tuples like
+        ``[(chess.E2, chess.E4)]``. An arrow from a square that is pointing to
+        the same square is drawn as a circle, like ``[(chess.E2, chess.E2)]``.
+    :param size: Adjust size of the SVG image by setting its width and
+        height, like ``(600, 600)``.
```
niklasf: *size* is supposed to by just one number, not a tuple, since the board is always quadratic.

chess/svg.py
```diff
@@ -130,10 +130,13 @@ def board(board=None, squares=None, flipped=False, coordinates=True, lastmove=No
...
+    :param style: Add CSS markup to change the default style sheet of the SVG
+        image.
```
niklasf: I like the original description better. CSS is not markup, stylesheet is a word.

chess/svg.py
PedanticHacker: Well, the docs here says `width and height`, so I figured you need to pass a tuple. Maybe someone wants to have a non-quadratic chessboard size because of space limitation perhaps. Your call on this one.

chess/svg.py
PedanticHacker: `:param size:` What does this parameter actually offer for the user of python-chess to achieve? Can (s)he limit the maximum size of the chessboard with this parameter by passing in an int like 600? Or can (s)he adjust the actual size of the chessboard by passing in a tuple like (600, 600) which corresponds to (width, height) respectively? This needs to be explained clearly.

`:param style:` Then maybe elaborate on this, because even I can't understand this explanation for `style`, and so others may probably not understand as well. Also, some concrete examples would be very beneficial here.

chess/svg.py
PedanticHacker: Thank you. ;)

PR #179, opened 2017-10-24
Okay, Niklas, this is my last attempt to fix the docs

PedanticHacker:
This patch is the second (and hopefully the last) attempt to correctly document some arguments of the `chess.svg.board` function.

Also, the patch fixes the value of the `DEFAULT_STYLE` constant so that the string is on one line only and is not triple quoted and that the ugly backslash can be removed. There's no point for a constant's value to be laid out so that something can be added to it. Remember, it's a constant.

Furthermore, where I found a docstring that said `SVG` I elaborated on this by saying `SVG image`.

### DIFF
--- a/chess/svg.py
+++ b/chess/svg.py
@@ -63,11 +63,7 @@ DEFAULT_COLORS = {
     "square light lastmove": "#cdd16a",
 }
 
-DEFAULT_STYLE = """\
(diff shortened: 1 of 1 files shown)


niklasf:
Thanks :)

PR #180, opened 2017-10-24
Some small tweaks

PedanticHacker:
In README.rst, `IPython notebook` is fixed to `Jupyter Notebook`,
because `IPython Notebook` is deprecated and will be gone in the future,
also capitalize the word `Notebook` because that's part of the official
name.

In `chess.svg`, some small tweaks are made.


niklasf:
:+1:

PR #188, opened 2017-10-27
Update uci.rst

PedanticHacker:
You can cherry-pick these fixes if you like, but I think I did a good job. ;)


niklasf:
Thanks, indeed. I cherry-picked anyway, for example leaving:
> It takes the command [future](https://en.wikipedia.org/wiki/Futures_and_promises) as a single argument.

PR #190, opened 2017-10-27
Update svg.rst

PedanticHacker:
The word "notebook" is "Notebook" (capitalized) when refering to the IPython/Jupyter Notebook web application, so I fixed that.

### DIFF
--- a/docs/svg.rst
+++ b/docs/svg.rst
@@ -2,7 +2,7 @@ SVG rendering
 =============
 
 The :mod:`chess.svg` module renders SVG images (mostly for IPython/Jupyter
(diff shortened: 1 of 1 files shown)


Mk-Chan:
From the logs it looks like it didn't actually fail it just hung up after finishing. Don't know what caused it.

PedanticHacker:
Yeah. I have a sence something is wrong on Travis CI side.

niklasf:
Yeah, it's hard to track down, because I could never reproduce the hangs locally. I'll just ignore test results that are obviously unrelated, until this is fixed.

PR #197, opened 2017-10-28
Update README.rst

PedanticHacker:
Those "doctest skip" and "doctest ellipsis" comments are not needed in the example code.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -25,7 +25,7 @@ and handling of common formats. This is the Scholar's mate in python-chess:
 
     >>> board = chess.Board()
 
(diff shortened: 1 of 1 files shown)


niklasf:
nope, they are needed. these examples are doctested to ensure they are working

PedanticHacker:
Yes, I know. But then those comments are then shown here: https://pypi.python.org/pypi/python-chess

The visitor of the PyPI website should not see those "doctest skip/ellipsis" comments when (s)he is looking at the example code. Can those comments be hidden when the README.rst is compiled for https://pypi.python.org/pypi/python-chess?

niklasf:
yes, done

PR #208, opened 2017-11-01
Some trivial API and code simplification

Mk-Chan:


niklasf:
Thank you, looks good.

PR #209, opened 2017-11-03
Code simplification and fixed bug in level command string building

Mk-Chan:


niklasf:
Nice, getting rid of a lot of repetition.

PR #210, opened 2017-11-03
Fixed a bunch of errors and warnings from pylint

Mk-Chan:
pylint is surprisingly useful! Just found it today. It seems to report errors, good practice, finds repetition and a bunch of other useful things. I analyzed the whole source with it and it rates python-chess terrible at 6.51/10 :p . Maybe make an issue of this?

### DIFF
--- a/chess/xboard.py
+++ b/chess/xboard.py
@@ -18,11 +18,12 @@
 # You should have received a copy of the GNU General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 
+import concurrent.futures
+import threading
+import random
+
 from chess.engine import EngineTerminatedException
 from chess.engine import EngineStateException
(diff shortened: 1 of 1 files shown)


LocutusOfPenguin:
there are some such tools like landscape.io and more.
To some degree i fixed found problems on my project, but to get around 100% is almost impossible.
For example: sometimes y forced to comment a (public) function, and the comment is longer as the function itself - like getters , ha

Mk-Chan:
Yes, I didn't even get the xboard module to 100% for this reason. It complains about stupid things like variable names even for very simple functions. But at least to some extent it will help good practice.

PedanticHacker:
This is a very welcome change. Yeah, the code is more readable this way. Thumbs up, Mk-Chan! ;)

PR #214, opened 2017-11-13
Added undo and remove API

Mk-Chan:


niklasf:
Yes, I guess so. As long as the outer command is queued it should be fine as is. Do you think there might be an issue with the worker pool size?

Mk-Chan:
I'll look into that, btw here's another instance of a failed 'passed' job https://travis-ci.org/Mk-Chan/python-chess/jobs/301406865

Mk-Chan:
Actually now that I think about it, why would the worker pool size be an issue? It's probably implemented as a list internally right? Should be highly extensible. Maybe I didn't exactly understand what you meant?

Issue #229, opened 2017-11-18
How to get the square on which a piece is on?


PedanticHacker:
Say I have a piece on a given file and rank and want to get the square that this piece is on. So I want to get the piece (which I know how to get) and then get the square of that piece as well. There's no function in python-chess to get the square of a given piece. How can I get this data? I need this data in my chess GUI to pass as the `tail` field for the `Arrow` namedtuple object. Niklas, please help me. Here's my code so far...

```python
#! /usr/bin/env python

"""
This module is the execution point of the chess GUI application
named Chess Titan.
"""

import sys

import chess
import chess.svg

from PyQt5.QtCore import pyqtSlot, Qt
from PyQt5.QtSvg import QSvgWidget
from PyQt5.QtWidgets import QApplication, QWidget


class MainWindow(QWidget):
    """
    Create a surface for the chessboard.
    """
    def __init__(self):
        """
        Initialize the chessboard.
        """
        super().__init__()

        self.setWindowTitle("Chess GUI")
        self.setGeometry(300, 300, 800, 800)

        self.widgetSvg = QSvgWidget(parent=self)
        self.widgetSvg.setGeometry(10, 10, 600, 600)

        self.boardSize = min(self.widgetSvg.width(),
                             self.widgetSvg.height())
        self.coordinates = True
        self.margin = 0.05 * self.boardSize if self.coordinates else 0
        self.squareSize = (self.boardSize - 2 * self.margin) / 8.0
        self.pieceToMove = [None, None]

        self.board = chess.Board()
        self.drawBoard(drawArrow=())

    @pyqtSlot(QWidget)
    def mousePressEvent(self, event):
        """
        Handle left mouse clicks and enable moving chess pieces by
        clicking on a chess piece and then the target square.

        Moves must be made according to the rules of chess because
        illegal moves are suppressed.
        """
        if event.x() <= self.boardSize and event.y() <= self.boardSize:
            if event.buttons() == Qt.LeftButton:
                if self.margin < event.x() < self.boardSize - self.margin and self.margin < event.y() < self.boardSize - self.margin:
                    self.file = int((event.x() - self.margin) / self.squareSize)
                    self.rank = 7 - int((event.y() - self.margin) / self.squareSize)
                    self.targetSquare = chess.square(self.file, self.rank)
                    self.piece = self.board.piece_at(self.targetSquare)
                    self.fileCharacter = chr(self.file + 97)
                    self.rankNumber = str(self.rank + 1)
[...]

PR #230, opened 2017-11-18
Add a `square()` method to `chess.Piece` class

PedanticHacker:
This might be useful and handy for someone. For me it certainly will be. I hope my implementation is correct.


niklasf:
Honestly I don't think this makes any sense. It's not using any information from piece. You might just as well call `chess.square(file, rank)` directly.

PedanticHacker:
Please look at my code in #229. I'm hyperventilating here.

Issue #232, opened 2017-11-18
`arrows` argument name is not correct


PedanticHacker:
Having an `arrows` argument of the `chess.svg.board()` function is not correct, because we can draw a circle with the same argument (and a circle is not an arrow). I therefore suggest we rename this argument to `markings` and accommodate other things.

niklasf:
I prefer `arrows` anyway.

PedanticHacker:
Maybe `indicators`? 🤔

PR #233, opened 2017-11-19
Update `chess/svg.py`

PedanticHacker:
Try to make the code more readable. Correct the wrongly named argument `arrows` of `chess.svg.board()` to `markings` and therefore also object `Arrow` to `Marking` + other things.

This relates to issue #232.

### DIFF
--- a/chess/svg.py
+++ b/chess/svg.py
@@ -58,8 +58,8 @@ DEFAULT_COLORS = {
 }
 
 
-class Arrow(collections.namedtuple("Arrow", "tail head")):
-    """Details of an arrow to be drawn."""
+class Marking(collections.namedtuple("Marking", "tail head")):
+    """Details of a marking to be drawn."""
 
     __slots__ = ()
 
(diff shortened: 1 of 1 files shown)


chess/svg.py
```diff
@@ -177,6 +178,7 @@ def board(board=None, squares=None, flipped=False, coordinates=True, lastmove=No
             "class": " ".join(cls),
             "stroke": "none",
             "fill": fill_color,
+            "class": "square",
```
niklasf: This would override the existing class, which already contains square (and possibly dark or light and/or lastmove) anyway.

chess/svg.py
```diff
@@ -259,21 +262,21 @@ def board(board=None, squares=None, flipped=False, coordinates=True, lastmove=No
...
+            head = []
+            head.append((xtip, ytip))
+            head.append((shaft_x + dy * 0.5 * marking_size / hypot,
+                         shaft_y - dx * 0.5 * marking_size / hypot))
+            head.append((shaft_x - dy * 0.5 * marking_size / hypot,
+                         shaft_y + dx * 0.5 * marking_size / hypot))
 
             ET.SubElement(svg, "polygon", {
-                "points": " ".join(str(x) + "," + str(y) for x, y in marker),
+                "points": " ".join(str(x) + "," + str(y) for (x, y) in head),
```
niklasf: I prefer tuple unpacking without extra parens.

chess/svg.py
```diff
@@ -259,21 +262,21 @@ def board(board=None, squares=None, flipped=False, coordinates=True, lastmove=No
                 "stroke-width": str(SQUARE_SIZE * 0.2),
                 "opacity": "0.5",
                 "stroke-linecap": "butt",
-                "class": "arrow",
+                "class": "tail",
```
niklasf: Head and tail intentionally have the same CSS class, so they can be referred to together. `line.arrow` is an easy way to select the shaft.

Issue #218, opened 2017-11-15
Xboard end states


Mk-Chan:
Xboard records the state of the program so we should provide the programmer some way to check whether the game has ended or not.

I know that one can access the `board` member and check for each end state (checkmate, 50-move, 75-move, threefold, stalemate, insufficient material). But in the case the engine chooses to resign or accept a draw (by agreement) `board` is not helpful. In this case it is simple, the Xboard class can provide a wrapper API to check for any kind of end state (the normal ones + draw and resign).

The problem I have is that through a contrived sequence, such a scenario can occur:
```
-> User calls go()
-> User waits for go() to end
-> Control returns to user
-> Engine offers draw
```
In such a case, how is it possible to inform the user that the engine has offered a draw? I don't think it's acceptable to expect the user to multi-thread to handle such a scenario by conditional waiting in one thread on a flag.

Mk-Chan:
Unless it's illegal to offer draw during the opponent's turn?

PedanticHacker:
Why is everyone talking about XBoard and/or Winboard? These are all CECP commands. What do these GUIs (XBoard and Winboard) have to do with CECP?

Mk-Chan:
Because CECP is the protocol Xboard/Winboard used therefore it is AKA the Winboard/Xboard protocol check it out here: http://hgm.nubati.net/CECP.html, http://hgm.nubati.net/newspecs.html, https://en.wikipedia.org/wiki/Chess_engine#Interface_protocol, etc...

niklasf:
If that's allowed (I have no idea) then it's indeed hard to fit into the request/response scheme. Perhaps another handler that's called from an engine thread? Unfortunately, like the info handler, it would make it easy for inexperienced users to run into data races. But I don't know what else we can do that's portable across Python 2 and  3 😞

@PedanticHacker All the XBoard vs. Winboard vs. CECP vs. "UCI is the future" add little to the dicussion of the issue at hand. Better make a new issue, if you've got a question.

Mk-Chan:
The exact line that caused this confusion is this:

```
And at the end of a search the engine can print the move it has decided to play, possibly accompanied by a draw offer, or it can decide to terminate the game.
```

From http://hgm.nubati.net/CECP.html
So the terminate the game will be a `result xyz` and the draw offer will be `offer draw`.
[...]

PR #234, opened 2017-11-19
Added offer_draw() API and DrawHandler

Mk-Chan:
So I just wanted to put this here for review. I will of course add formal tests soon(with MockProcess probably) but for now I have used a modified test engine to give `offer draw` at appropriate times to test the sample script (plus few variations for each case). The cases tested were:
1. Engine offers draw during it's turn (engine gives `offer draw` after every pv post)
2. Engine offers draw during opponent's turn (engine gives `offer draw` after `move xyz`)
3. Human offers draw during engine's turn (engine gives `offer draw` after receiving `offer draw`)

This is the way I imagined a user would handle it. Please share your thoughts.
The script was:

```
import chess.xboard
import time
import logging

logging.basicConfig(level=logging.DEBUG)

class MyPostHandler(chess.xboard.PostHandler):
    def post_info(self):
        print(self.post)
        super(MyPostHandler, self).post_info()

post_handler = MyPostHandler()
draw_handler = chess.xboard.DrawHandler()

engine = chess.xboard.popen_engine("./wyldchess")
engine.post_handlers.append(post_handler)
engine.draw_handler = draw_handler

engine.xboard()
engine.st(10)
engine.go(async_callback=True)

# To test 3. just send engine.offer_draw() and then check engine.end_result as follows:
"""
time.sleep(5)
engine.offer_draw()
if engine.end_result == chess.xboard.DRAW:
    print("GOT RESULT: " + str(engine.end_result))
"""

# Value greater than 10 tests 2.
# Value less than 10 tests 1.
time.sleep(11)
if draw_handler.pending_offer:
    print("GOT DRAW OFFER")
    engine._end_game(chess.xboard.DRAW) # Replace with engine.result() once result() function added
print("GOT RESULT: " + str(engine.end_result))

print("DONE!")
```

### DIFF
--- a/chess/xboard.py
+++ b/chess/xboard.py
@@ -296,7 +296,7 @@ class Engine(object):
         self.author = None
         self.features = FeatureMap()
         self.pong = threading.Event()
-        self.ping_num = None
+        self.ping_num = 123
         self.pong_received = threading.Condition()
         self.auto_force = False
         self.in_force = False
@@ -536,7 +536,6 @@ class Engine(object):
         def command():
             with self.semaphore:
                 with self.pong_received:
-                    self.ping_num = random.randint(1, 100)
                     self.send_line("ping " + str(self.ping_num))
                     self.pong_received.wait()
 
--- a/chess/xboard.py
+++ b/chess/xboard.py
@@ -34,6 +34,89 @@ from chess.engine import _spur_spawn_engine
 import chess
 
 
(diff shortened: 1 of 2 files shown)


Mk-Chan:
Refer to #218

Mk-Chan:
Draw handler should be generalized to say, ResultHandler and we can handle both resignation and draw within in.

Mk-Chan:
So I've added proper tests now as well

niklasf:
Thank you! Alright ... let's go with this.

PR #235, opened 2017-11-19
Add documentation for unicode() and _repr_svg_()

PedanticHacker:
This patch adds documentation for the unicode() and _repr_svg_() methods. It's useful for people reading the documentation to know that these two handy functions are available and what they do.


niklasf:
Thank you, but these are undocumented intentionally. `_repr_svg_` because calling it directly is unusual, `unicode` because I want to be able to change it without warning.

PR #240, opened 2017-11-22
Update xboard.py

PedanticHacker:
I went through all 1252 lines of code and fixed the grammar and orthography + added a default value of `False` to the `ponder` kwarg of the `pondering` method under the `Engine` class, because the `ponder` kwarg was described as being `False` by default but really wasn't set to nothing, so I fixed that as well.

### DIFF
--- a/chess/xboard.py
+++ b/chess/xboard.py
@@ -506,7 +506,7 @@ class Engine(object):
             return future
         else:
             # Avoid calling future.result() without a timeout.
-            # In Python 2, such a call can not be interrupted.
+            # In Python 2, such a call can't be interrupted.
             while True:
                 try:
                     return future.result(timeout=FUTURE_POLL_TIMEOUT)
--- a/chess/xboard.py
+++ b/chess/xboard.py
@@ -77,7 +77,7 @@ class DrawHandler(object):
 
     def pre_offer(self):
         """
-        Received a new draw offer about to be processed.
+        Processes the newly received draw offer.
 
         When subclassing, remember to call this method of the parent class in
         order to keep the locking in tact.
@@ -86,7 +86,7 @@ class DrawHandler(object):
 
     def post_offer(self):
         """
-        Processing of a draw offer has been finished.
+        Finishes processing of the newly received draw offer.
 
         When subclassing, remember to call this method of the parent class in
         order to keep the locking in tact.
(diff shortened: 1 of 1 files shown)


PedanticHacker:
Maybe "Received a score in centipawns." is even better if "Receives a score in centipawns."?

ddugovic:
My point was more that an engine has at most one score at any given time (except if there are multiple PVs, but I don't know if that's a feature of CECP/XBoard).

PedanticHacker:
Thank you, mister Dugović, for your tip. It made me revisit my patch and now I've done some more fixes. Please take a look and report back.

chess/xboard.py
```diff
@@ -133,20 +133,20 @@ class PostHandler(object):
...
     >>> post_handler.post["score"]
     34
 
     See :attr:`~chess.xboard.PostHandler.post` for a way to access this dictionary
     in a thread-safe way during search.
 
-    If you want to be notified whenever new information is available
-    you would usually subclass the *InfoHandler* class:
+    If you want to be notified whenever new information is available,
+    you would usually subclass the :class:`~chess.xboard.InfoHandler` class:
```
Mk-Chan: Since you're doing these changes, could you also make it `chess.xboard.PostHandler

chess/xboard.py
```diff
@@ -572,14 +572,14 @@ def draw(self, async_callback=None):
         command = self.command("draw")
         return self._queue_command(command, async_callback)
 
-    def pondering(self, ponder, async_callback=None):
+    def pondering(self, ponder=False, async_callback=None):
```
Mk-Chan: I'm not so sure about this because this default argument will imply that when someone calls `pondering()` it will in fact be turned off which seems un-intuitive. If anything, it should be `True`. In fact now that I think about it it should probably be renamed to `ponder()` since it's not the actual Xboard spec (which has `hard` and `easy`)

Mk-Chan:
Thanks for the grammatical fixes!
About the conversion from things like `Stop engine` to `Stops engine...`, I had written in that way to maintain uniformity with the UCI module(which I'm sure I haven't been very successful in either :p). So you'll have to change that too if this is to work.

niklasf:
Thanks, I'll wait for @Mk-Chan to sign off the new changes and then merge. (Actually `cannot` is usually preffered over `can not`. Was `def pondering` -> `def ponder` intended?)

Mk-Chan:
The `cannot` is preferred yes, and `pondering` -> `ponder` should be done if putting that default argument because `engine.ponder()` makes more sense than the alternative.

It doesn't really matter that much since it's not technically part of the spec. I had just made it as a convenience because pondering being controlled by `easy` and `hard` is not very transparent.

PedanticHacker:
Sorry about the `cannot` to `can not` fix.

Mk-Chan:
No problem looks good to me at least, thanks: :)

PR #244, opened 2017-11-27
Fix a comment in chess/pgn.py

PedanticHacker:
Fix a comment that doesn't seem right. The grammar style should match all other comments.

### DIFF
--- a/chess/pgn.py
+++ b/chess/pgn.py
@@ -877,8 +877,8 @@ def read_game(handle, Visitor=GameModelCreator):
             line = handle.readline()
             continue
 
-        # An empty line means the end of a game. But gracefully try to find
-        # at least some content if we didn't even see headers so far.
+        # An empty line means the end of a game. Try to gracefully find
+        # at least some content if the headers aren't seen so far.
         if found_game and line.isspace():
             visitor.end_game()
             return visitor.result()


niklasf:
thanks. i think it more clearly states what's going on as is

PedanticHacker:
I know, but saying "we didn't even see" is kinda odd to address the reader of the code. Nowhere in code did I find such style. Please fix it. :)

niklasf:
I think it's ok, as long as it gets the point across. Not worth your time to try and rewrite them all:
```
chess/__init__.py-        if self.stack:
chess/__init__.py:            # Castling rights do not change in a game, so we can assume them to
chess/__init__.py-            # be filtered already.
--
chess/__init__.py-    def _castling_uncovers_rank_attack(self, rook_bb, king_to):
chess/__init__.py:        # Test the special case where we castle and our rook shielded us from
chess/__init__.py-        # an attack, so castling would be into check.
--
chess/gaviota.py-            else:
chess/gaviota.py:                # Else LZMA86. We have to build a fake header.
chess/gaviota.py-                DICTIONARY_SIZE = 4096
--
chess/syzygy.py-
chess/syzygy.py:        # If en passant is not possible, we are done.
chess/syzygy.py-        if not board.ep_square or self.variant.captures_compulsory:
--
chess/syzygy.py-            elif v == 0:
chess/syzygy.py:                # If there is not at least one legal non-en-passant move we are
chess/syzygy.py-                # forced to play the losing en passant cature.
--
chess/uci.py-            # The ponder move must be legal after the bestmove. Generally, we
chess/uci.py:            # trust the engine on this. But we still have to convert
chess/uci.py-            # non-UCI_Chess960 castling moves.
--
chess/xboard.py-        def command():
chess/xboard.py:            # Use the join(builder) once we parse usermove=1 feature
chess/xboard.py-            move_str = " ".join(builder)
--
test.py-        # with a series of legal moves from one of the 960 starting positions.
test.py:        # Decision: We don't care. Neither does Stockfish or lichess.org.
test.py-        fen = "1r1k1r2/5p2/8/8/8/8/3N4/R5KR b KQkq - 0 1"
```

PR #247, opened 2017-11-30
Added analyze and exit

Mk-Chan:
I'm not exactly sure how to test this formally, it does work with a simple script though. I just subclassed `PostHandler` to print the PV and sent the program on a `engine.analyze(async_callback=True)` and stopped it 10 seconds later with `engine.exit()` successfully.


niklasf:
Thank you!

PR #250, opened 2017-12-06
Update xboard.py

PedanticHacker:
There were some imperfections made in the last patch for this module. I fixed that.


Mk-Chan:
That timing! I was just about to comment about the *feature egt*

PedanticHacker:
I am very pedantic. I always strive for perfection. This is what I do best. 😉

PR #251, opened 2017-12-07
Xboard commands and doc fixes

Mk-Chan:


chess/xboard.py
```diff
@@ -919,7 +960,7 @@ def egtpath(self, egt_type, egt_path, async_callback=None):
 
         :return: Nothing.
         """
-        if not (egt_type in self.features.get("egt")):
+        if not egt_type in self.features.get("egt"):
```
niklasf: btw. even more elegant is `not in`

Mk-Chan: Haha yes, C-style habits die hard

Issue #249, opened 2017-12-06
Travis stalling tests


Mk-Chan:
Occasionally Travis CI stalls some tests after completing the following script line successfully:
```
if [[ $PERFT -ne 1 ]]; then coverage run --source chess test.py -vv; fi
```

It fails to retrieve the exit code which should be printed as follows:
```
The command "if [[ $PERFT -ne 1 ]]; then coverage run --source chess test.py -vv; fi" exited with 0.
```

It says instead:
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

This indicates to me that some test case did not actually `TearDown` properly or something. Since we tear down the Stockfish and Crafty tests with `terminate` which probably translates to SIGTERM, maybe we should instead SIGKILL it if SIGTERM fails for 5 seconds. Or even just SIGKILL directly

Mk-Chan:
I think we can test this by adding a print or something before we wait for the `is_terminated` condition variable and after it terminates at the lowest level(in engine.py) then push a bunch of spurious commits to a temporary branch. I haven't been able to reproduce it on my machine either so it's a bit hard to test. What do you think?

niklasf:
Restarted tests a couple of times after https://github.com/niklasf/python-chess/compare/7b47775be7590bb61405a51b2e1cae8d89ec9ccf...5eada1c799efda05b049bf2f309a490368443a10. So far I haven't seen further stalls.

niklasf:
Just seen it again, so it's only rarer at best.

niklasf:
It's been a while since the last stalled test, but it just happened again: https://travis-ci.org/niklasf/python-chess/jobs/419535950

Now that the tests are split we can see that it was hanging after `coverage run --source chess --append -m doctest README.rst`.

Mk-Chan:
I don't get it, the doctest searches for the
```
>>> <python stuff>
<output>
```
lines and makes sure the <output> matches the actual output for <python stuff> right? Why would that be an intermittent bug?

niklasf:
No idea. Maybe it's `coveralls`, so the next command, that's hanging. Added an `echo` in between, so we will definitely know if the doctests exited.

niklasf:
https://travis-ci.org/niklasf/python-chess/jobs/462928149

niklasf:
https://travis-ci.org/niklasf/python-chess/jobs/465300963 (UciEngineTestCase)

Mk-Chan:
[...]

PR #252, opened 2017-12-10
Xboard option parsing and tests

Mk-Chan:


niklasf:
Thanks.

Once we figured out #249 I wanna to remove all `time.sleep` calls. Those can never be required if everything else is alright.

Mk-Chan:
Actually the sleep calls are there after the async function calls to pretty much guarantee execution of that function before the next main thread one.

PR #263, opened 2018-02-20
force utf-8 encoding

dannygale:
Force utf-8 encoding when reading from README.rst to avoid ascii encoding problems


niklasf:
Thanks!

PR #270, opened 2018-04-20
Fix handling multiple features in on line

gbtami:
xboard protocol enables two from of feature command:
feature NAME=VALUE
feature NAME1=VALUE1 NAME2=VALUE2 ...

In latter case str.split() doesn't save values surrounded by double quotes. This patch fixes this.
It also handles the case where an option feature is not in the first feature in the feature list.

### DIFF
--- a/chess/xboard.py
+++ b/chess/xboard.py
@@ -347,13 +347,11 @@ class Engine(object):
             raise EngineStateException("Engine produced an error: {}".format(err_msg))
 
         command_and_args = buf.split()
-        if not command_and_args:
+        if command_and_args[0] == "#" or not command_and_args:
             return
 
         if len(command_and_args) == 1:
-            if command_and_args[0] == "#":
-                pass
-            elif command_and_args[0] == "resign":
+            if command_and_args[0] == "resign":
                 return self._resign()
         elif len(command_and_args) == 2:
             if command_and_args[0] == "pong":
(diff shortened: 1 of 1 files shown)


Mk-Chan:
This is great, @ZirconiumX informed me of this yesterday. Hopefully this helps xboard boots use this: https://github.com/careless25/lichess-bot

gbtami:
@Mk-Chan yep, https://lichess.org/@/PyChessBot can play standard chess with this patch now.
Next step will be to implement "variant" command.

chess/xboard.py
```diff
@@ -346,13 +347,11 @@ def on_line_received(self, buf):
             raise EngineStateException("Engine produced an error: {}".format(err_msg))
 
         command_and_args = buf.split()
-        if not command_and_args:
+        if command_and_args[0] == "#" or not command_and_args:
```
Mk-Chan: Ideally match this PR: #271

niklasf:
Hi, thanks for the patches. I just merged #271 and cherry-picked 08ba755b987405e9bbc0b65fc6dcfe6cf6dee684.

ZirconiumX:
@niklasf Would you mind getting a PyPI release out so that things like careless25/lichess-bot can put them in their requirements.txt?

niklasf:
Sure. I'll wait for Travis and then make a release.

niklasf:
Published in [v0.23.2](https://github.com/niklasf/python-chess/releases/tag/v0.23.2).

PR #274, opened 2018-04-23
ebf

flok99:
EBF is not an official status-line component but some UCI programs emit it.


niklasf:
Sorry, what's ebf?

Mk-Chan:
It's a measurement of branching factor.

niklasf:
Ah, thanks. I was thinking maybe it has something to do with the book format :)

Seems simple enough to add/maintain. Let's do it.

flok99:
Cool thanks

Op do 26 apr. 2018 18:52 schreef Niklas Fiekas <notifications@github.com>:

PR #277, opened 2018-04-29
Added pondering parsing support

Mk-Chan:
The PostHandler should now seamlessly parse the pondering output from the engine in compliance with the spec (I hope). I have tested it with a simple script of mine checking with Crafty that both the ponderhit and non-ponderhit seem to work well.

I will write formal tests for this after some human review.

### DIFF
--- a/chess/xboard.py
+++ b/chess/xboard.py
@@ -38,6 +38,18 @@ DUMMY_RESPONSES = [ENGINE_RESIGN, GAME_DRAW] = [-1, -2]
 RESULTS = [WHITE_WIN, BLACK_WIN, DRAW] = ["1-0", "0-1", "1/2-1/2"]
 
 
+def try_move(board, move):
+    try:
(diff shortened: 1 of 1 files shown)


niklasf:
Thank you. Looks good.

Mk-Chan:
I'm having some difficulty writing the test for this. The issue I'm having is that in order to set (and therefore, test) the ponder move, we check for `move_received.is_set()`. However, there is no check to ensure we have recorded the ponder move before we end up testing for it.

In reality, there is not really any point detecting when we have the ponder move set I think because you can't really control the engine and you can't wait on it since the engine is not _required_ to ponder even when ponder is on.

This sounds like that DrawHandler issue all over again if you want to  set an `on_ponder_move_received` or something. On the other hand, while the engine is pondering, we are only expected to quit, make a move and *maybe* reset the board with `new` or `setboard` which we can do anyways.

PR #278, opened 2018-04-29
BaseBoard handles full FENs

Antek55:
BaseBoard._set_board_fen crashed when passed a full FEN, e.g. from Board.fen(), since it only used the position part. Now it drops the latter part and sets the BaseBoard as expected.

### DIFF
--- a/chess/__init__.py
+++ b/chess/__init__.py
@@ -841,6 +841,7 @@ class BaseBoard(object):
 
     def _set_board_fen(self, fen):
         # Ensure the FEN is valid.
+        fen = fen.split(' ', 1)[0]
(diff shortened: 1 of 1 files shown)


niklasf:
Hi, thanks for the patch. I am not sure that this is the right thing to do. Perhaps instead it should raise a `ValueError` with a more descriptive error message? If the user really intends to drop the other FEN parts, they could easily do `.split(' ', 1)[0]` themselves to explicitly acknowledge this, or get a `board_fen()` in the first place.

Mk-Chan:
I think what should be changed there(if you must change something) is the function name to something like set_pieces_on_squares

PR #288, opened 2018-05-22
Generalize the LC0 8-move requirement

Mk-Chan:
Pretty trivial change I think

### DIFF
--- a/chess/uci.py
+++ b/chess/uci.py
@@ -815,7 +815,7 @@ class Engine(object):
 
         return self._queue_command(command, async_callback)
 
-    def position(self, board, async_callback=None):
+    def position(self, board, lookback=0, async_callback=None):
         """
         Set up a given position.
 
@@ -826,6 +826,9 @@ class Engine(object):
         *ucinewgame* command before the *position* command.
 
         :param board: A *chess.Board*.
+        :param lookback: An integer defining how many moves back should the
+                         initial FEN and the moves leading up to the current
+                         FEN be. This can override the repetition check.
 
         :return: Nothing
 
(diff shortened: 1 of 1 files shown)


Mk-Chan:
Something just failed for seemingly no reason...
https://travis-ci.org/niklasf/python-chess/jobs/382288829#L1847

Mk-Chan:
Ok, looking seriously at this, https://syzygy-tables.info/gaviota is 404. My entire Travis CI test suite failed because of this: https://travis-ci.org/Mk-Chan/python-chess/jobs/382287956

I don't think this is python-chess' fault but it may be cause for concern for the test suite if this 404 is permanent.

Mk-Chan:
I suppose it's not specified by UCI but neither is the lc0 requirement. In fact UCI defines a position as the initial fen + all moves following. I think this option also provides a way to achieve that (by giving it the total stack size as the argument).

The default could be changed for the convenience of lc0 sure... But that seems a bit partial. Although I'm not against supporting the latest innovation in computer chess, I'm not so sure ethically choosing 8 as default is fine.

Either way, it's your choice. If I were to go with the lc0 requirement I'd prefer being able to get however many moves I want back (since lc0 is still in development)

niklasf:
You can convinced me that hardcoding a value that's useful specifically for a single engine is not ideal (even if it does not significantly hurt other engines).

But instead of making this configurable, I'll try to fix the underlying performance issue and always send all moves. In the meantime I'll still release the status-quo, together with some bugfixes.

Mk-Chan:
So what exactly is the performance issue? Can we not just store a list of all strings and just print " ".join(list) it?

niklasf:
Yes, that will be the goal. Currently all moves have to be replayed to retrieve the initial FEN and normalize notation of castling moves.

niklasf:
Now sending the full history again, but more efficiently. Via: 9fd2d3e7e5619e306cc5cf0c64971ec6619f95bc, ae48918d21a923a153f00e49b01d726343de9a94 and 446280c1174a8c57ed5ed4dc8df92c92e5755a2a.

Issue #231, opened 2017-11-18
Get captured pieces


PedanticHacker:
Is there a (simple) way in python-chess to get captured pieces? I want to display an SVG image for each captured piece.

Mk-Chan:
For every piece type(pawn,knight...queen) just popcount the bitboard(in BaseBoard) and subtract from the maximum number of that piece type.

For variants that modify the number of pieces you'll just have to count how many of each piece.

PedanticHacker:
Thanks. 😉

PR #296, opened 2018-07-08
Counting captured pieces

Nicals:
I have implemented issue #231.

This method does not count the absolute number of captured pieces, but the difference between the ones on the board and the one that should initially be on the board (see test case).

I've seen that Scid vs PC uses this technique, so I hope it is good enough.

### DIFF
--- a/chess/__init__.py
+++ b/chess/__init__.py
@@ -1246,6 +1246,28 @@ class BaseBoard:
         board.set_chess960_pos(sharnagl)
         return board
 
+    def count_captured(self, piece, color):
+        """
+        Get the number of captured pieces on the current position.
+
+        :param piece: Type of the piece to count (chess.ROOK, ...)
+        :param color: Color of the piece to check (chess.WHITE or chess.BLACK)
+
+        :raises: :exc:`ValueError` if the piece type is invalid.
+        """
+        if piece == QUEEN:
+            piece_count = 1
+        elif piece in (ROOK, BISHOP, KNIGHT):
+            piece_count = 2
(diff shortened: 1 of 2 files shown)


niklasf:
Mhh ... and even then it would be more of a material difference, than just captures, because of promotions.

Mk-Chan:
That's true. Then how about a histogram of captures by piece type updated after every move if it was a capture?

Nicals:
A histogram of capture piece should be another option, but it is a different issue.
We cannot always rely on this because if I load a game position without the move history, I won't be able to obtain any valuable information.

I think, both of these method should be provided. This one as a rough capture evaluation, and the capture histogram for a more precise one if the state of the board allows it.

I suggest renaming the method *estimate_capture_count*.
We should open an issue to discuss the capture histogram since it solves another thing.
What do you think ?

Mk-Chan:
But if you start at a position without history, you cannot, by definition, say that any captures have been played because it is indistinguishable from a starting position.

I guess we need to define the use case more clearly.

ddugovic:
I don't understand why captured pieces would be more interesting than material difference (in particular, if a passed pawn is about to be captured, why we care what it promoted to prior to being captured).

Mk-Chan:
I think material difference is fine, it's just not actually a count of captured pieces. The function name does not express intent if it's returning material diff

Mk-Chan:
Looking at the issue as quoted by OP again, it seems the requirement is difference of material from some startpos. In which case I guess it's ok

PR #300, opened 2018-07-25
Update setup.py

MerlinSp:
Hey! Thanks for letting us use your great work. I would suggest you change the name of the module in setup.py to "pythonchess". This way we would be able to pip install it and import the module without being in the main directory. Import statements with a "-" won't work. Thanks, and keep the good work up!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -72,7 +72,7 @@ def extra_dependencies():
 
 
 setuptools.setup(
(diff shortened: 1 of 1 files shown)


Mk-Chan:
I don't think there even is a python-chess module. The modules are chess, chess.uci, chess.pgn etc

PR #307, opened 2018-10-09
Adding type aliases for Color, PieceType, and Square

coreylowman:
# Motivation

Python 3 added type hinting to allow for more explicit type annotations to be added to python code. Right now colors, piece types, and squares are all raw types (bool, int, and int respectively), which doesn't lend itself to type hinting:

```python
def do_something(color: bool, piece_type: int, square: int):
    ...
```

If you aren't familiar with the python-chess api, you don't know what the values for the parameters mean, could be, or come from.

# Changes

Three [type aliases](https://docs.python.org/3.6/library/typing.html#type-aliases) were added to help this. This makes the above example:

```python
def do_something(color: chess.Color, piece_type: chess.PieceType, square: chess.Square):
    ...
```

This is much more explicit about where the values are coming from.

These additions have no effect on existing code, but does require python 3.5+.

### DIFF
--- a/chess/__init__.py
+++ b/chess/__init__.py
@@ -38,10 +38,12 @@ import struct
 
 COLORS = [WHITE, BLACK] = [True, False]
 COLOR_NAMES = ["black", "white"]
+Color = bool
 
 PIECE_TYPES = [PAWN, KNIGHT, BISHOP, ROOK, QUEEN, KING] = range(1, 7)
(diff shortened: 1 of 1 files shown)


niklasf:
Thanks. There's no way to provide typings for the existing methods without dropping 3.4 support, right?

coreylowman:
That's correct, type annotations were added in 3.5. I think there was a way to do it before 3.5, but I'm not sure how it worked and I don't think its compatible with the 3.5+ way.

PR #308, opened 2018-10-10
Update svg.rst

PedanticHacker:
Added the word 'licenses' at the end of the sentence because it finishes it; without it that sentence is kinda dull.


niklasf:
Thanks. Imo these are better as is.

PR #313, opened 2018-10-23
Fix SQUARE_SIZE constant to 50

PedanticHacker:
This constant is currently wrong. The default size of the SVG chessboard is 400, divide that with 8 (as we have 8 files/ranks) and the result is 50, so the square size is 50. I have noticed in my app that it is off for exactly 5 pixels. Now I finally found the culprit. Please compile a new version of python-chess.

### DIFF
--- a/chess/svg.py
+++ b/chess/svg.py
@@ -27,7 +27,7 @@ import math
 import xml.etree.ElementTree as ET
 
 
-SQUARE_SIZE = 45
+SQUARE_SIZE = 50
 MARGIN = 20
 
 PIECES = {


niklasf:
I think this is based on a misunderstanding.

There is no default size in `chess.svg`, so the values for `SQUARE_SIZE` and `MARGIN` can never be "wrong".

The viewbox size is calculated as `8 * SQUARE_SIZE + 2 * margin`. With the default margin of 20 that is 400, yes. Without a margin the size of the viewbox is only 360.

If you explicitly specify `chess.svg(..., size=400, coordinates=False)` then the internal viewbox is still 360 and `SQUARE_SIZE = 45` is still correct in that coordinate system. Some manual conversion is needed between the viewbox coordinate system and display coordinates.

niklasf:
> There is no default size in chess.svg, ...

Btw. this is only true since April (74aed6dca0df9f9dfd0f800b655414c207167073). Maybe that's the origin of this confusion.

PedanticHacker:
> Some manual conversion is needed between the viewbox coordinate system and display coordinates.
Like how should I do that? How can I do this manual conversion?

niklasf:
Rule of three:

square size in display coordinates = `chess.svg.SQUARE_SIZE` * image size in display coordinates / viewbox size.

So for example:

50 = 45 * 400 / 360

PedanticHacker:
Thank you! This is what I need.

PedanticHacker:
When I check the dimensions of my SVG widget with `print(self.size())` (where `self` is reffering to the `QSvgWidget` object), the result I get is always 640 × 480, even though I set the geometry of my window to 800 × 800 and I also set the `size` argument of `chess.svg.board` to `800`. What gives? Any ideas?

PR #314, opened 2018-10-23
Fix SQUARE_SIZE constant to 50

PedanticHacker:
This constant is currently wrong. The default size of the SVG chessboard is 400, divide that with 8 (as we have 8 files/ranks) and the result is 50, so the square size is 50. I have noticed in my app that it is off for exactly 5 pixels. Now I finally found the culprit. Please compile a new version of python-chess.

### DIFF
--- a/chess/svg.py
+++ b/chess/svg.py
@@ -27,7 +27,7 @@ import math
 import xml.etree.ElementTree as ET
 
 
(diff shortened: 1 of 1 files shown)


niklasf:
Please so my previous answer on https://github.com/niklasf/python-chess/pull/313#issuecomment-432331030.

PR #324, opened 2018-11-22
Enable support for reversed(GameNode) built-in

### LINKED ISSUE #319: Add support for the reversed() built-in on appropriate iterators
https://docs.python.org/3/library/functions.html#reversed
```
reversed(seq)

    Return a reverse iterator. seq must be an object which has 
a __reversed__() method or supports the sequence protocol 
(the __len__() method and the __getitem__() method with integer 
arguments starting at 0).
```
As discussed offline, I'm planning on implementing support for this built-in with a PR when I can get the time.

Discussion in #95 might be relevant.

rpdelaney:
Closes #319 

I didn't see anything obvious to add this to except a GameNode. It looks like variations are just lists of moves so they are already iterators. Let me know if I missed something and I'll throw it in, now that I've worked out how to do this.

As mentioned in the `TODO` I added to the code, this doesn't support slices yet. I don't really need that in `python-chess-annotator` so I went ahead and submitted this as-is. If you're interested in that feature I can open a separate issue to get that together later.

### DIFF
--- a/chess/pgn.py
+++ b/chess/pgn.py
@@ -388,6 +388,19 @@ class GameNode:
     def __str__(self):
         return self.accept(StringExporter(columns=None))
 
+    def __len__(self):
+        """
+        Return the length of the main variation
+        """
+        node = self
+        length = 1
+
+        while node.variations:
+            node = node.variations[0]
+            length += 1
+
+        return length
+
 
 class Game(GameNode):
     """
--- a/chess/pgn.py
+++ b/chess/pgn.py
@@ -508,7 +508,7 @@ class Headers(collections.MutableMapping):
 
     def is_wild(self):
         # http://www.freechess.org/Help/HelpFiles/wild.html
-        wild = self.get("Variant", "").lower() in [
+        wild = self.get("Variant", "").lower() in [  # noqa E514
             "wild/0", "wild/1", "wild/2", "wild/3", "wild/4", "wild/5",
             "wild/6", "wild/7", "wild/8", "wild/8a"]
 
--- a/chess/pgn.py
+++ b/chess/pgn.py
@@ -401,6 +401,32 @@ class GameNode:
 
         return length
 
+    def __getitem__(self, key):
+        """
+        Get the 'key'th child node of this GameNode
+
+        TODO: add support for slices
+        """
+        try:
+            assert type(key) is int
+        except AssertionError:
+            raise TypeError  # index must be integer type
+        try:
+            assert key >= 0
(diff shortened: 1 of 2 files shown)


rpdelaney:
Looking at this again, I think `len(Game)` will return 1 on a game with no moves played. That's probably wrong.

niklasf:
Thanks. Will have a look at this soon.

rpdelaney:
I guess it is weird to have this and not `next() ` and `iter()` too.

```
Having seen the mechanics behind the iterator protocol, it is easy to add iterator behavior to your classes. Define an __iter__() method which returns an object with a __next__() method. If the class defines __next__(), then __iter__() can just return self:

class Reverse:
    """Iterator for looping over a sequence backwards."""
    def __init__(self, data):
        self.data = data
        self.index = len(data)

    def __iter__(self):
        return self

    def __next__(self):
        if self.index == 0:
            raise StopIteration
        self.index = self.index - 1
        return self.data[self.index]
```
I'll throw this in sometime in the next few days.

PedanticHacker:
The list type has a method reverse(), which reverses all the items in that list in-place. Why would you even want that functionality in python-chess to be built in? If anyone, for some reason, wants a list of moves in the reversed order, (s)he can just call reverse() on the resulting list. Say, my_list_of_moves.reverse().

rpdelaney:
@PedanticHacker It's important to reverse the order when performing engine analysis so that the hash table will be populated with future positions as we go along. As I mentioned in my previous comment, I want to add support for the linear iterator methods as well --  I had tunnel vision on the reversed order because I need it for my chess annotator.

Regardless, it doesn't look like a `GameNode` has a list method as it is either, so I'm not sure how your advice helps us without making further changes to this module.

PedanticHacker:
The way I am imagining to implement this is:

```
class GameNode:
    .
    .
    .
    def __reversed__(self):
        yield something
```

And then one would be able to use the reversed() built-in function on an instance of the GameNode object.

`reversed(instance)`

rpdelaney:
@PedanticHacker Providing a `__reversed__` method would also get this done. But per the docs, adding a `len()` and the `__getitem__()` method is enough to get `reversed()` off the ground via the sequence protocol.

However, as I mentioned above, my implementation of `__len__()` as it stands in this PR might be wrong. I need to find time to confirm that (and fix, if necessary) before this is merged.

@niklasf Do you think it would be better to simply implement a `__list__()` method? Would that obviate this PR?

PR #330, opened 2018-11-26
Remove the TODO

PedanticHacker:
Now that we have and use BB_EMPTY (instead of BB_VOID), we can safely remove this TODO tag.


niklasf:
The intent was to keep this around for another release or two.

PR #332, opened 2018-11-28
Fix Headers.is_wild() always returning None

rpdelaney:
Maybe I'm mistaken but it looks like this function is going to return None no matter what and not have any side-effects either.

Since it looks like `Headers.variant()` method (the only place where the `.is_wild()` method is called) expects a boolean value, I made this change.

I'm not sure where a regression test could be added to `test.py` but I'd like to include one. Let me know where you think it belongs and I'll add it.

### DIFF
--- a/chess/pgn.py
+++ b/chess/pgn.py
@@ -508,7 +508,7 @@ class Headers(collections.MutableMapping):
 
     def is_wild(self):
         # http://www.freechess.org/Help/HelpFiles/wild.html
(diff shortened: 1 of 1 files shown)


niklasf:
Thanks. I noticed this in your coding style fixes and already committed de2a657b67aa3a784044417ca4614cd14b17cbfb and also a regression test (fd7d6fa29c8a5e4784866ea656440da5954ba316).

PR #336, opened 2018-12-03
Fix some grammar

PedanticHacker:
Just fixed some grammar, no code changes here.


niklasf:
Thanks. Since this comes up repeatedly: I prefer that documentation refers to `self` as "the game", "the move", etc. rather than just "a game", "a move".

PR #344, opened 2019-01-02
Use PIECE_SYMBOLS in _set_board_fen

gbtami:
My goal is to use Fairy-Stockfish with lichess-bot in a new web app. This change enables extending PIECE_SYMBOLS with new letters used in other variants FEN like Makruk, Sittuyin and similar.


niklasf:
Thanks, this is a good change anyway.

PR #425, opened 2019-06-18
Rename variable 't' to something more descriptive

PedanticHacker:
I've renamed the variable 't', which is very undescriptive. Every variable should be properly named so that its name alone describes why is it there and what data does it hold. I hope I understand the code enough to give the variable 't' a proper name. If not, please rename it yourself.

### DIFF
--- a/chess/svg.py
+++ b/chess/svg.py
@@ -88,15 +88,15 @@ def _svg(viewbox: int, size: Optional[int]) -> ET.Element:
 
 
 def _text(content: str, x: int, y: int, width: int, height: int) -> ET.Element:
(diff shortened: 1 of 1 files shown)


niklasf:
No thanks. It's the only variable there, so there's no chance to confuse it with anything else.

PedanticHacker:
A good, descriptive name for a variable is always a good idea. Someone reading the code might ask “What does that ‘t’ stand for?” Just like I am asking now. Can you at least give some proper name for it?

PedanticHacker:
@niklasf, what does that 't' stand for anyway? 🤔

PR #428, opened 2019-07-18
add time

SotirisKantz:
Hi, decided to add a new class named Time on python-chess/chess/__init__.py. With the function get_time(), it calculates the current running time of the game and prints it.


SotirisKantz:
Running

niklasf:
Thanks. This seems more like a general utility, not necessarily even chess related.

PR #441, opened 2019-10-01
New pinners function

kraktus:
Creation of the "pinners" function that takes a color and a square and return the set of pinners squares from the given square.

If the given square is empty or not pinned to the king of the given color it will return an empty set of squares.

It is a combination of attackers, pin, and is_pinned

An useful function that completes well the pin and is_pinned function.


Mk-Chan:
A possible faster implementation is to use pseudo rook (for queen and rook) and pseudo bishop (for queen and bishop) attacks from the king of the given color. Pseudo attacks means not accounting for blockers.

However you have to be careful if you consider an x-ray pinner through your own piece as a pinner or not

niklasf:
Is it true that there will be at most one pinner per square?

kraktus:
Absolutely, a pinned piece can only have one Pinner at a time. In fact the `pin_mack` function already finds the pinners, but I didn't find an elegant way to extract this information.

Mk-Chan:
I just noticed... What is this .DS_Store binary?

kraktus:
An folder for saving preferences on Mac OS: https://en.wikipedia.org/wiki/.DS_Store Anyway, I'm pretty new to Github but this PR looks like a real mess, I'll try again later when I'll be more confortable with it.

PR #476, opened 2020-03-12
Update engine.rst

PedanticHacker:
Shouldn’t this be written like it’s proposed in this patch? Or am I missing something?


niklasf:
Thanks, there was an inconsitency. Fixed via 331e88ab6f9bcb3b28293e59e56f05cac91ad70e.

The `#` is common notation for "mate in ...".

PR #478, opened 2020-03-13
Properly tweak engine scores

PedanticHacker:
I think the way the engine scores should be presented is this.


niklasf:
The sign is important fot mate scores, and the other scores are centi-pawn (`Cp`) values.

PR #479, opened 2020-03-13
Fix grammar

PedanticHacker:
These docstrings were fixed to accomodate to the wording style of other docstrings throughout python-chess. The documentation is therefore fixed.

### DIFF
--- a/chess/engine.py
+++ b/chess/engine.py
@@ -281,7 +281,7 @@ class Option(NamedTuple):
 
 
 class Limit:
-    """Search termination condition."""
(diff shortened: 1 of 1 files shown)


chess/engine.py
```diff
@@ -281,7 +281,7 @@ def is_managed(self) -> bool:
 
 
 class Limit:
-    """Search termination condition."""
+    """Searches termination condition."""
```
niklasf: Looks good to me except this one. `Limit` does not search anything, it *is* the search termination codition.

PR #480, opened 2020-03-14
Simplify release.py module

PedanticHacker:
Call the int() function just once. Current code calls the int() function twice in the case  master is behind.

Furthermore, the shebang line is improved, visually and functionally. Calling ‘env’ is preffered. It ensures that the Python interpreter is found (by the Linux ‘env’ command) even if installed on some custom path. This shebang line needs to change in every module that exists for the python-chess project. I can make a PR for this.

### DIFF
--- a/release.py
+++ b/release.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#! /usr/bin/env python3
 # Helper script to create and publish a new python-chess release.
 
 import os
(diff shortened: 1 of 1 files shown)


release.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#! /usr/bin/env python3
```
niklasf: Skipping this change, because of the extra space and because `release.py` is not intended to run in a venv.

PR #481, opened 2020-03-14
Remove the obsolete first line

PedanticHacker:
Now that python-chess supports only Python 3.6+ (for some time
now), we don’t need this first line anymore. Python 3 does not need
this line because UTF-8 is universal. In any other module in the python-chess project it
is obsolete. I am removing it in this module. Other modules should follow suit.


niklasf:
Thanks! Removed them all, except in `setup.py`, which should always be able to run on Python 2.

PR #482, opened 2020-03-14
Prettify the ‘coding’ line

PedanticHacker:
The ‘coding’ line can be written like this as well. I think this way looks prettier.


niklasf:
No, thanks. The current style has no disadvantages, and some minor advantages (editor support).

PR #485, opened 2020-03-17
Simplify linter ignore marks

PedanticHacker:
Apparently, you don’t need to specify
`# noqa: E501`
to tell the linter to ignore a line as being too long; it suffices to specify only
`# noqa`
and the linter won’t complain either.

Looks better and shortens already too long lines, albeit each line for 6 characters only.

Also, if you ever decide to switch to a different linter, hardcoding the error code is not a good idea, because different linters have different error codes for "line too long".


PedanticHacker:
I am not sure where to report this particular documentation bug, but the `# doctest: +SKIP` marks are seen in the python-chess documentation. These marks should not have been seen. I've noticed 2 occurences in the **SVG rendering** section. There may be other occurences elsewhere in the docs as well.

niklasf:
No, thanks. I prefer to silence only specific errors. In the hypothetical scenario that I switch linters and the code is no longer E501, it will show an error for the overlong lines. So there's no possibility to forget about this.

PR #486, opened 2020-03-17
Bad Sphinx rendering

PedanticHacker:
This Sphinx code
.. py:data:: chess.WHITE
    :annotation: = True
is rendered as
chess.WHITE= True
if you look closely (there’s no space in front of the equal sign, separating ‘E’ and the equal sign.

Can this be fixed or is this not doable? Discard this PR if you find this not solvable.

### DIFF
--- a/docs/core.rst
+++ b/docs/core.rst
@@ -7,7 +7,7 @@ Colors
 Constants for the side to move or the color of a piece.
 
 .. py:data:: chess.WHITE
-    :annotation: = True
+    :annotation:  = True
 
 .. py:data:: chess.BLACK
     :annotation: = False


niklasf:
I see the issue, but the suggested change does not have any effect :disappointed:

niklasf:
It appears to be this issue with the specific CSS/theme on readthedocs.org: https://github.com/readthedocs/sphinx_rtd_theme/issues/694

```css
.rst-content dl:not(.docutils) .property {
    display: inline-block; /* space lost here! */
    padding-right: 8px;
}
```

breathe67:
I have just started using Sphinx & Breathe (version as of Jun 1, 2020) to convert my Doxygen docs to the RTD theme. I'm also having this problem, where my C++ types and other identifiers have no spaces between them.

PedanticHacker:
Someone should report this bug upstream.

PR #487, opened 2020-03-17
"Seven Tag Roster", and not "7 tag roster"

PedanticHacker:
According to [Wikipedia](https://en.wikipedia.org/wiki/Portable_Game_Notation#Tag_pairs) and also Chess.com and other credible sites, the seven PGN tag pairs are reffered to and written as the **Seven Tag Roster**, or STR for short, and not as **7 tag roster** (as in the pgn.py module). Also, you can't speak of the STR as being default. STR is just STR. So I am fixing this, which will fix the documentation text.

### DIFF
--- a/chess/pgn.py
+++ b/chess/pgn.py
@@ -527,7 +527,7 @@ class Game(GameNode):
 
     @classmethod
     def without_tag_roster(cls: Type[GameT]) -> GameT:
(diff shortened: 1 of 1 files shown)


niklasf:
Fixed. Default refers to the values of the STR.

PR #489, opened 2020-03-17
Add a period to block comment

PedanticHacker:
Added a period at the end of a block comment. Every block comment ends with a period and so should this one.


niklasf:
Don't think this makes sense.

PR #494, opened 2020-03-18
Simplify unittest.skipIf() decorator condition

PedanticHacker:
This change provides the same result, but it is more readable. Also, someone on StackOverflow ensured that checking for the OS version is more reliable by using platform.system(): https://stackoverflow.com/questions/1854/what-os-am-i-running-on/58071295#58071295

Furthermore, the current Python version comparison is the same now and not complicated. It's simpler to compare the Python version as being less than version 3.8.1 as opposed to less than or equal to 3.8.0 and less than 3.8.1.

### DIFF
--- a/test.py
+++ b/test.py
@@ -2591,7 +2591,7 @@ class PgnTestCase(unittest.TestCase):
         self.assertTrue(isinstance(node, MyGameNode))
 
 
-@unittest.skipIf(sys.platform == "win32" and (3, 8, 0) <= sys.version_info < (3, 8, 1), "https://bugs.python.org/issue34679")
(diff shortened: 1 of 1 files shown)


niklasf:
No, this change is not equivalent:

```python
>>> "3.10.1" < "3.8.1"
True
>>> (3, 10, 1) < (3, 8, 1)
False
```

PedanticHacker:
Hmmm, I see your point. But this is okay, though:
```
"3.10.1" < "3.08.1"
False
```
But, of course, we can't compare Python versions like this.

Should I open a new Python bug report? Maybe there's a bug lurking in the Python interpreter.

niklasf:
Everything is working as expected. Strings are compared lexicographically.

PR #503, opened 2020-03-20
Fix example code + 1 typo

PedanticHacker:
Ensure that the file is closed afterwards (by using the 'with' statement) and use list comprehension (that's an optimization, because list comprehensions are very fast).

### DIFF
--- a/chess/pgn.py
+++ b/chess/pgn.py
@@ -1382,21 +1382,18 @@ def read_headers(handle: TextIO) -> Optional[Headers]:
 
     >>> import chess.pgn
     >>>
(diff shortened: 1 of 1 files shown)


niklasf:
Did you test this code?

PedanticHacker:
My PR code has a syntax error. Oops! Now I've tried the list comprehension like this:
`kasparov_offsets = [offset for offset in pgn if "Kasparov" in headers.get("White", "?")]`
I hope that's correct?

Using the 'with' statement in the terminal is a dumb idea, I admit, because the file gets closed and you get `ValueError: I/O operation on closed file`. Sorry about that.

Issue #500, opened 2020-03-20
Update unicode representation of chess board.


sepiolGodess:
Right now the representation of chessboard in Unicode is:

```
♜ ♞ ♝ ♛ ♚ ♝ ♞ ♜
♟ ♟ ♟ ♟ ♟ ♟ ♟ ♟
. . . . . . . .
. . . . . . . .
. . . . . . . .
. . . . . . . .
♙ ♙ ♙ ♙ ♙ ♙ ♙ ♙
♖ ♘ ♗ ♕ ♔ ♗ ♘ ♖
```

It has irregular spacing in the squares with no pieces, update this function 
`chess/__init__.py/unicode`
to use this Unicode symbol ⭘ for empty squares, better lining :)

```
♜ ♞ ♝ ♛ ♚ ♝ ♞ ♜
♟ ♟ ♟ ♟ ♟ ♟ ♟ ♟
⭘ ⭘ ⭘ ⭘ ⭘ ⭘ ⭘ ⭘
⭘ ⭘ ⭘ ⭘ ⭘ ⭘ ⭘ ⭘
⭘ ⭘ ⭘ ⭘ ⭘ ⭘ ⭘ ⭘
⭘ ⭘ ⭘ ⭘ ⭘ ⭘ ⭘ ⭘
♙ ♙ ♙ ♙ ♙ ♙ ♙ ♙
♖ ♘ ♗ ♕ ♔ ♗ ♘ ♖
```

sepiolGodess:
@niklasf solved this issue in the ecb97060c3e0e6eb86cd439e9a51f51c1f7a9777 commit.

PR #504, opened 2020-03-20
Drop the 'u' for the Unicode string

PedanticHacker:
Since we dropped Python 2 support, I think the u”·” should only be “·”, i.e., without the letter ‘u’, since all strings in Python 3 are now inherently Unicode.

Plus some small tweaks.

### DIFF
--- a/chess/__init__.py
+++ b/chess/__init__.py
@@ -1147,7 +1147,7 @@ class BaseBoard:
         Useful for pretty-printing to a terminal.
 
         :param invert_color: Invert color of the Unicode pieces.
-        :param borders: Show borders and a coordinate margin.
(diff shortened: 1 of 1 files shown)


chess/__init__.py
```diff
@@ -1172,7 +1172,7 @@ def unicode(self, *, invert_color: bool = False, borders: bool = False) -> str:
                 if piece:
                     builder.append(piece.unicode_symbol(invert_color=invert_color))
                 else:
-                    builder.append(u"·")
+                    builder.append("·")
```
Shibe: Maybe you can also replace the the `·` with `⭘` as mentioned in [this issue]( https://github.com/niklasf/python-chess/issues/500).

PR #523, opened 2020-03-25
Stockfish version 11

PedanticHacker:
Now we have Stockfish 11. Maybe we should update to use its 11th version, although the fake one?


niklasf:
Usually the fuzz corpus should be expanded, not edited. Added some Stockfish 11 output (0b523fc11973d142f59da29f86c37bccb8bfe0f1).

PR #525, opened 2020-03-27
Carry-Rippler?!

PedanticHacker:
There’s no Carry-Rippler trick. There’s only ripple carry, found in electronics.


PedanticHacker:
I guess I missed something. Please fix everything from _carry_rippler to _ripple_carry.

PR #526, opened 2020-03-27
Carry-Rippler?!

PedanticHacker:
Fixed the code and tests for the wrongly named _carry_rippler() function. It is _ripple_carry() now. In electronics, there is the ripple-carry adder. Carry-Ripler does not exist anywhere. I don't know who named this function like that. Anyway, it is fixed now.

Would be a good idea to notify about this function rename in the changelog (if the pull request is merged).


niklasf:
https://www.chessprogramming.org/Traversing_Subsets_of_a_Set#All_Subsets_of_any_Set

PR #527, opened 2020-04-01
Maybe we missed this one?

PedanticHacker:
When we changed the interpoint to a circle, we may have missed this one. Not related to the unicode() method, but still.

This was a normal period character. Now I changed it to an interpoint, because I don't think the circle character applies here. Or does it? @niklasf, please check it out.


niklasf:
Thanks for checking. This one is intentionally ASCII.

PR #529, opened 2020-04-10
Fix typing errors in gaviota

GeorgeAntono:


GeorgeAntono:
Ignore my second commit, it's identical to the first. It was an  accidental duplicate commit.

PedanticHacker:
And how does that fix the typing? It basically removes it. Was that intended?

niklasf:
Thanks for raising this. I don't like the potential that this has to silence previously fatal usage of uninitialized fields, so I tried a slightly different patch.

PR #542, opened 2020-04-27
Add a new method is_promoting() to Board

PedanticHacker:
I've written a new method for the Board class that checks whether a white/black pawn move is a promotion attempt. It is fully legal. If it needs to be pseudo-legal, let me know and I'll rewrite it.

@niklasf, please write a test for this method (if you decide to commit it).

### DIFF
--- a/chess/__init__.py
+++ b/chess/__init__.py
@@ -2916,6 +2916,29 @@ class Board(BaseBoard):
         touched = BB_SQUARES[move.from_square] ^ BB_SQUARES[move.to_square]
         return bool(touched & self.pawns or touched & self.occupied_co[not self.turn] or move.drop == PAWN)
 
+    def is_promoting(self, move: Move) -> bool:
+        """
+        Checks whether the *move* is either a white pawn or a black pawn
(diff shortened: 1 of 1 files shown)


chess/__init__.py
```diff
@@ -2916,6 +2916,29 @@ def is_zeroing(self, move: Move) -> bool:
...
+        """
+        Checks whether the *move* is either a white pawn or a black pawn
+        attempting to promote.
+
+        Returns ``True`` only if all conditions are met. These conditions
+        include that the moving piece is actually a pawn that is attempting to
+        jump from the pre-last to the last rank and is not pinned to its king.
+        Also, the target square must be vacant or a legal capture.
+        """
+        if piece != PAWN:
```
zhaoshaojun, line 2929: we need the following line at the beginning of the function, right?
`piece = self.piece_type_at(move.from_square)`
otherwise, piece at line 2929 is not defined.

PedanticHacker:
piece = self.piece_type_at(move.from_square)

It was a total mishap I didn’t include this line.

PR #545, opened 2020-04-28
Add a new method is_promoting() to Board

PedanticHacker:
I've written (and now fixed firstly the code and then the indentation errors) a new method for the Board class that checks whether a white/black pawn move is a promotion attempt. It is fully legal. If it needs to be pseudo-legal, let me know and I'll rewrite it.

@niklasf, please write a test for this method (if you decide to commit it).

### DIFF
--- a/chess/__init__.py
+++ b/chess/__init__.py
@@ -2916,6 +2916,31 @@ class Board(BaseBoard):
         touched = BB_SQUARES[move.from_square] ^ BB_SQUARES[move.to_square]
         return bool(touched & self.pawns or touched & self.occupied_co[not self.turn] or move.drop == PAWN)
 
+    def is_promoting(self, move: Move) -> bool:
+        """
+        Checks whether the *move* is either a white pawn or a black pawn attempting
+        to promote.
+
+        Returns ``True`` only if all conditions are met. These conditions include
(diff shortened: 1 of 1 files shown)


niklasf:
Thanks. I agree that something like this is needed. Will review soonish.

PedanticHacker:
I desperately needed a way in my application to test whether a move is a promotion attempt and I had to write my own function. Having it in python-chess would be beneficial to other people. So in the spirit of open source, I am giving something back to the community. :)

@niklasf, I hope you like my approach `if move.uci() not in [move.uci()[0:4] for move in self.legal_moves]`.

This method was tested for all possible positions that can occur in a standard chess game, and it was always returning the correct boolean value. Among other tests, I've tested if the promotion move is attempted when the pawn is pinned to its king, and the function was correct at returning False, because that move would be illegal. All logic is correct.

If you'd like the pseudo-legal version of is_promoting(), let me know. Also, I'm bad at writing tests (there are none in my application), so I'm leaving this to you. ;)

niklasf:
I implemented something slightly more general (but also more efficient). Usage for this purpose:

```python
if board.find_move(chess.B7, chess.B8).promotion:
    # ...
```

Issue #541, opened 2020-04-27
We should start using Stockfish 11


PedanticHacker:
https://github.com/niklasf/python-chess/blob/ad5a4388e311624b8add4bbbf21aa46485703afc/appveyor.yml#L11

The line of code I am referencing, and the corresponding ones below it, should be fixed to use Stockfish 11.

niklasf:
via #549

PR #549, opened 2020-04-29
Fix spelling error and introduce Stockfish 11 instead of Stockfish 10

TesseractA:
As the title comments; hopefully I've not neglected anything that has "Stockfish 10" in it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ before_install:
      wget https://stockfish.s3.amazonaws.com/stockfish-11-linux.zip
      unzip stockfish-11-linux.zip
      mkdir -p bin
-     cp stockfish-11-linux/Linux/stockfish_20011801_x64_modern bin/stockfish
+     cp stockfish-11-linux/Linux/stockfish_20011801_x64_bmi2 bin/stockfish
      chmod +x bin/stockfish
      export PATH="`pwd`/bin:${PATH}"
      which stockfish || (echo $PATH && false)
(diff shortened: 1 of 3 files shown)


TesseractA:
Fixes #541 and my conscience about spelling errors

niklasf:
Thanks. Can you please update Stockfish `.travis.yml` as well?

PedanticHacker:
I think it doesn't hurt if we use stockfish_20011801_x64_bmi2.exe. The BMI2 version is much much faster. Intel processors since 2013 (basically from the Haswell processor design onwards) all use this new instruction set. The BMI2 instruction set (i.e., the pext and pdep instructions) makes it possible to speed up Stockfish calculations quite a bit, actually.

TesseractA:
@PedanticHacker How about AMD processors? (Also, my desktop is a 2011 computer, yes outdated, but probably cannot run bmi2)

TesseractA:
Well I'm not installing stockfish with pychess anytime soon. But from my perspective it looks like this one is done.

niklasf:
Thanks!

This is just the automated test suite, so performance doesn't really matter. Both modern or bmi2 are fine, if the testing infracstructure supports it.

brianprichardson:
If including SF here then certainly the lowest common denominator version should be used.  Many have PCs without BMI2 support (some even without popcnt).  Python-chess should not be triggering problems with SF which is not part of this repo, IMHO.  If you want to leave it as is fine.  If it triggers a bunch of "python-chess does not work for me anymore" issues it is easy to see if this is the cause and revert at that point.

TesseractA:
Well, what we could do is test whether stockfish installs and runs. If not, then we can create an assert that causes us to install another version.

PR #551, opened 2020-04-29
Fix the coordinate layout

PedanticHacker:
I've fixed my coordinates layout problem. Use this code and test whether you get the same accurate layout of the coordinate strips, especially the bottom one (where the letters b, d, f and h should now **not** overlap the chessboard).


niklasf:
`height // 1.5` is not the proper way to find the center

PedanticHacker:
But it works for me. Does it not work for you?

The thing is that you need some spacing between the bottom edge of the chessboard and the letters in the bottom coordinate strip. Letters `a`, `c`, `e` and `g` are okay, but letters `b`, `d`, `f` and `h` have this protruding vertical line (especially `f`) where the protrusion is such that it penetrates the edge of the chessboard.

PR #558, opened 2020-05-16
Update engine.py with .clear()

PedanticHacker:
As for the __init__.py, this line in engine.py can also lose the “del” keyword and can be replaced with the .clear() function/method.


PedanticHacker:
Hmmm, maybe the reference must be deleted (destroyed) and not just the value cleared. That’s why Travis build failed?

PR #569, opened 2020-06-10
Fix default value from () to []

PedanticHacker:
The default value of the `arrows` argument of the `chess.svg.board()` function can not be an empty tuple, it must be an empty list, since the `arrows` argument accepts a **list of tuples**. So, I think, we should fix that.


niklasf:
I was trying to avoid the [mutable default argument trap](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments), but I think it's fine, because the argument is clearly intended to be read-only.

PR #583, opened 2020-06-29
Added clock_comment method to the GameNode class

jesseadamrandallhester:
I hope that I've created this pull request in the proper way! :)

There are other ways to accomplish what we want, i.e., allowing access to timestamps. We could have a more general method that returns a dict whose keys are commands (from https://www.enpassant.dk/chess/palview/enhancedpgn.htm) and whose corresponding values are the corresponding parameters. The method defined in this PR simply searches a node's comment for the "clk" command in particular and returns its parameter as a string (or "" if the "clk" comment is not present), which seems like the simplest and least opinionated way to make timestamps available. However, I am not attached to this particular way of doing it and I'm open to feedback if we want to do it differently.

### DIFF
--- a/chess/pgn.py
+++ b/chess/pgn.py
@@ -106,6 +106,8 @@ MOVETEXT_REGEX = re.compile(r"""
 
 SKIP_MOVETEXT_REGEX = re.compile(r""";|\{|\}""")
 
+CLOCK_REGEX = re.compile("\[%clk\s([^\]]*)\]") # used by the clock_comment method of GameNode
+
 TAG_ROSTER = ["Event", "Site", "Date", "Round", "White", "Black", "Result"]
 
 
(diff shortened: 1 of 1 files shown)


jesseadamrandallhester:
Hmm. It looks like Travis doesn't approve of how the method uses try/except as control flow:

>chess/pgn.py:468: error: Item "None" of "Optional[Match[str]]" has no attribute "groups"

I've tested the method with PGNs that contain timestamps and PGNs that don't, and when you create a game from a PGN that doesn't have timestamps, calling the method on one of the nodes just returns the empty string.

Travis also found a large number of errors in syzygy.py, which should be the same as the version in the master branch. I don't know what to make of these errors.

jesseadamrandallhester:
Besides the errors in syzygy.py, this could be improved by defining the comment regex outside of the method so that it isn't recompiled every time the method is called. It would also probably be better not to catch all exceptions and just return the empty string, because that could suppress unrelated exceptions. I'll make a couple of tweaks and update the branch.

niklasf:
Thanks. I'll need a while to think about this.

PR #597, opened 2020-08-13
Aditionally -> Additionally (+ something else)

PedanticHacker:
The sentence "The operations ``bm`` and ``bm`` also accept a list of legal moves in the current position." is a little bit odd, especially the "``bm`` and ``bm``" part.


niklasf:
Thanks. That was supposed to be *am* and *bm*.

PR #600, opened 2020-08-15
Fix docstrings

PedanticHacker:
Fixed some docstrings.


chess/__init__.py
```diff
@@ -3660,25 +3660,25 @@ def __len__(self) -> int:
     # MutableSet
 
     def add(self, square: Square) -> None:
-        """Adds a square to the set."""
+        """Adds a square to a set."""
```
niklasf, line 3663: This refers to *the* self object, not just any.

PedanticHacker: Ah, I see. So don't apply this particular change.

chess/__init__.py
```diff
@@ -3664,21 +3664,21 @@ def add(self, square: Square) -> None:
         self.mask |= BB_SQUARES[square]
 
     def discard(self, square: Square) -> None:
-        """Discards a square from the set."""
+        """Discards a square from a set."""
```
niklasf, line 3667: Same here (and maybe more).

PedanticHacker: Everything is fixed in the 3rd commit.

PR #601, opened 2020-08-15
Fix docstrings (second attempt)

PedanticHacker:
Finally fixed docstrings as they should be. Sorry for making a new PR, I got lost in all those branches and commits. Anyway, @niklasf , I hope you like my commit now.


niklasf:
Thanks.

PR #604, opened 2020-08-18
Fix some docstrings (new found)

PedanticHacker:
Fixed some newly found docstrings that needed some polishing.


niklasf:
*neither side has sufficient* is **not** equivalent to *either side has insufficient*

PR #606, opened 2020-08-18
fmvc -> fmvn

PedanticHacker:
The name of the field "fmvc" is wrong. Fixed that to the correct name "fmvn".


niklasf:
Thanks for the patches!

PR #607, opened 2020-08-18
Add a comma

PedanticHacker:
I believe there should be a comma. This was fixed in a PR before, but was rejected altogether with something I didn't fix properly. I hope you'll fix this now.


niklasf:
I don't think there should be a comma when describing logical consequences (https://english.stackexchange.com/a/30468).

PR #610, opened 2020-08-23
Fix some peculiarities in the __init__

PedanticHacker:
Fixed some peculiarities I found in the `__init__`. As I noticed throughout the code, python-chess does not use a comma before `and` or `or`, so I fixed those occurences and have tweaked one of the sentences a bit to be more understandable.

Also, the `STARTING_FEN` and `STARTING_BOARD_FEN` constants have a triple-quoted string on a line below them. That's kind of strange. Looks like a docstring, but it actually isn't. I converted that into a regular comment for each constant, also tweaked the sentences a bit.

Then another thing I did was fixed the docstring of the `Board()` subclass.

@niklasf, I hope you like my ideas.

### DIFF
--- a/chess/__init__.py
+++ b/chess/__init__.py
@@ -68,11 +68,11 @@ FILE_NAMES = ["a", "b", "c", "d", "e", "f", "g", "h"]
 
 RANK_NAMES = ["1", "2", "3", "4", "5", "6", "7", "8"]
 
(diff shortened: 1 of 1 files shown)


niklasf:
* Sphinx understands these "docstrings".
* If anything, I'd probably use Oxford commas everywhere.

Issue #624, opened 2020-09-07
Why not allow string square as a parameter to piece_type_at / color_at / piece_at?


tanmay2004:
While looking for a function which could tell me which piece was at a particular square, I found this: [piece_type_at](https://python-chess.readthedocs.io/en/latest/core.html#chess.BaseBoard.piece_type_at). I saw that it takes in an integer as input, which was very confusing because I didn't know what integer represented which square. After digging for a bit, I discovered it took integers from 0 to 63 representing each square from "a1" to "h8".

Then, I found the `chess.SQUARE_NAMES` constant. I used this to solve my problem, by first getting the index of my string square from it and then passing it as the parameter to `piece_type_at`. My question is: why not allow as input the square name, as a string? We could keep both options available by adding a parameter to the function. If the string is passed, we could easily get it's integer value. As I said, this can be easily accomplished outside of the module too, the user could get the index of the square and then pass it to `piece_type_at`, like I did. But, why not save him/her the trouble of having to figure it out?

I would be happy to make the changes and create a pull request :)

niklasf:
Allowing both for every function that works with square indexes will be quite messy and a considerable slowdown. Let's improve the documentation instead. Perhaps adding a function like `chess.parse_square(name: str) -> Square` would be simplest?

tanmay2004:
Oh, yeah that's true. `parse_square` seems like a good idea! Also, because a method to get the square as a string from it's integer value exists ([square_name](https://python-chess.readthedocs.io/en/latest/core.html#chess.square_name)), having the opposite available as well does make sense.

PR #625, opened 2020-09-07
Added parse_square function: #624

tanmay2004:


niklasf:
Thanks. I think it would be better to raise `ValueError` if this fails. Like https://github.com/niklasf/python-chess/blob/1e6353792df58b31c264ae3cfe8277ea8d579fb1/chess/__init__.py#L490.

tanmay2004:
If we used `.index` instead of `.find` as I have done, then it would automatically raise a `ValueError`. Is that what you mean?

tanmay2004:
@niklasf I'm confused, do you mean I should just return as I am right now, but with `.index`? Or should I use a try and catch and raise the `ValueError` with a custom message?

tanmay2004:
Wait, if @PedanticHacker made another pull request with the same changes, what should I do? Sorry, but I'm fairly new to contributing on Github.

PedanticHacker:
Don't worry, I just made the function a little bit tidy and I used the index() function so that a ValueError is raised if the function fails. Kudos are all to you, @tanmay2004! ;)

tanmay2004:
Thanks! Actually, I had just edited my branch with the `.index` change and was gonna commit it when I saw your pull request. So, I'm a bit confused....

niklasf:
Thank you, both.

PR #637, opened 2020-09-24
Fix chess.svg module

PedanticHacker:
- [x] Fix the docstring of `chess.svg.board()`.
- [ ] Add `DEFAULT_COLORS` to argument `colors` of `chess.svg.board()`.

### DIFF
--- a/chess/svg.py
+++ b/chess/svg.py
@@ -180,7 +180,7 @@ def board(board: Optional[chess.BaseBoard] = None, *,
         square is drawn as a circle, like ``[(chess.E2, chess.E2)]``.
     :param size: The size of the image in pixels (e.g., ``400`` for a 400 by
         400 board) or ``None`` (the default) for no size limit.
-    :param colors: Dictionary to override default colors. Possible keys are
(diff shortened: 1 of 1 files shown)


PedanticHacker:
My chess GUI, after updating python-chess to 1.0.0, displays arrows like in the image below, and it now doesn't display a circle at all. Everything worked in the version prior to 1.0.0. So, what gives? I thought changing the default value from `{}` to `DEFAULT_COLORS` of the argument `colors` of `chess.svg.board()` would fix the problem, but I was wrong. Anyway, take a look at this...
image
Also, passing `chess.svg.DEFAULT_COLORS` to the argument `colors` has the same effect as opposed to pass my own values.

niklasf:
`{}` is intended, because in order to allow partial overrides we need to check `DEFAULT_COLORS` anyway. The fallback is implemented here:

https://github.com/niklasf/python-chess/blob/014544bcb95a9e5c6fe8096119522802f1b08530/chess/svg.py#L119-L120

Issue #638, opened 2020-09-24
Colors and objects not appearing, or strange behavior, v1.0.0 only


PedanticHacker:
My chess GUI, after updating python-chess to 1.0.0, displays arrows like in the image below, and it now doesn't display a circle at all. Everything worked in the version prior to 1.0.0. So, what gives? I thought changing the default value from `{}` to `DEFAULT_COLORS` of the argument `colors` of `chess.svg.board()` would fix the problem, but I was wrong. Anyway, take a look at this...
![image](https://user-images.githubusercontent.com/14254085/94179359-b009b180-fe9c-11ea-9e30-2702488b52cb.png)
Also, passing `chess.svg.DEFAULT_COLORS` to the argument `colors` has the same effect as opposed to passing my own values.

_Originally posted by @PedanticHacker in https://github.com/niklasf/python-chess/pull/637#issuecomment-698487132_

niklasf:
Mhh ... so what code exactly are you using to produce the image, now?

PedanticHacker:
https://github.com/PedanticHacker/ChessFighter
Take a look at the code in my GitHub repo. Using the same version of PyQt5 as before.

PedanticHacker:
https://github.com/PedanticHacker/ChessFighter/blob/74c5d54719b01764df8491aad9ae22fdfebd19d6/source/window.py#L444
This line may be important. @niklasf, please investigate if I must change anything to accomodate v1.0.0. It's all so strange since it worked well in v0.31.4.

niklasf:
Ah, it's because SVG Tiny (and Qt's renderer) cannot deal with the new default color `#15781B80` which includes its opacity. I'll implement a workaround.

PedanticHacker:
The closest websafe color of `#15781B80` is `#006666`.

PedanticHacker:
So I fixed my issue. `"arrow green": "#006666"` fixes the issue for me.

![image](https://user-images.githubusercontent.com/14254085/94186705-4e9b1000-fea7-11ea-8311-922eb2cfcc7f.png)
It doesn't look as cool as it would with some opacity, but I'm happy. The important thing is to only use default colors that are websafe.

@niklasf, but if you manage to make a workaround, that'd be awesome!

PedanticHacker:
Now since `python-chess v1.0.1` is out, I will test with the default color `#15781B80` again. I just have to wait for `pip` to catch up.

niklasf:
Published.

[...]

PR #639, opened 2020-09-24
Change default color of "arrow green"

PedanticHacker:
Rather use the websafe color #006666 as a default for "arrow green", so that SVG renderers that can't render opacity (**looking at you, Qt's renderer!**) will be happy. This PR refers to issue #638.


niklasf:
Implemented support for opacity instead.

PedanticHacker:
Just saw. Let me try now on v1.0.1.

PR #641, opened 2020-09-24
Improve chess.svg.board() docstring

PedanticHacker:
Explain that regular color names like `red`, `green`, `blue`, and the like, can also be passed in as `color` arguments to `chess.svg.board()`.


niklasf:
I'd rather not guarantee that this will work indefinitely.

PR #644, opened 2020-09-25
Fix the shebang line

PedanticHacker:
Since we're not supporting Python 2.x anymore, reflect that by fixing the shebang line.


PedanticHacker:
Every occurence of the shebang line in `python-chess` needs to be fixed.

PR #646, opened 2020-09-25
Allow zeros in SAN castling notation

tanmay2004:
I believe that the correct way, or at least widely accepted way of writing the castle move (long and short) is with zeros. That's the way I generally write it too. So, inside the `parse_san` function, I've added replacing all zeros in the `san` string with 'O', so that it doesn't throw an error but instead recognizes the move as castling.

I couldn't really find any reference where the SAN notation for castling is defined. But, it is definitely used by many. I guess both with 'O' and with '0' is valid. I understand that `parse_san` will still return the parsed move with 'O's in it, but I think that's fine as other functions like `xboard` are doing the same.

P.S. I have no idea why it is showing 20 commits. I just did one. I tried to update my fork by pulling the main repo's commits into mine and then I added my thing. I think that essentially, only one file should be different (`__init__.py`) but I am very confused, sorry. Oh no, one of the checks failed. Does anybody know how I can fix this?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,40 @@
 Changelog for python-chess
 ==========================
 
+New in v1.0.0
+-------------
+
+Changes:
+
+* Now requires Python 3.7+.
+* `chess.engine` will now cut off illegal principal variations at the first
+  illegal move instead of discarding them entirely.
+* `chess.engine.EngineProtocol` renamed to `chess.engine.Protocol`.
+* `chess.engne.Option` is no longer a named tuple.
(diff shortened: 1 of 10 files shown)


PedanticHacker:
I'd suggest you close this PR, then delete your python-chess repo, and fork the python-chess repo again. Then do your edits in your freshly forked repo, and make a PR. It should work. Well, it must! ;)

tanmay2004:
Ahh okay, I guess I'll do that. But, could you please tell me how I can keep all my forks upto date or point me to a tutorial regarding the same? Thanks!

PR #647, opened 2020-09-25
Allow zeros in SAN castling notation 

tanmay2004:
Refer #646 for context. Same changes have been made here....

### DIFF
--- a/chess/__init__.py
+++ b/chess/__init__.py
@@ -2802,6 +2802,7 @@ class Board(BaseBoard):
         """
         # Castling.
         try:
+            san = san.replace('0', 'O')
             if san in ["O-O", "O-O+", "O-O#"]:
                 return next(move for move in self.generate_castling_moves() if self.is_kingside_castling(move))
             elif san in ["O-O-O", "O-O-O+", "O-O-O#"]:
--- a/chess/__init__.py
+++ b/chess/__init__.py
@@ -2802,10 +2802,9 @@ class Board(BaseBoard):
         """
         # Castling.
         try:
-            san = san.replace('0', 'O')
-            if san in ["O-O", "O-O+", "O-O#"]:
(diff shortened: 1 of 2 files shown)


PedanticHacker:
Hmmm, Travis is not happy.

tanmay2004:
Yeahh, why might they be failing? I just added a replace to the code and as far as I know, that particular `.replace()` method can't throw any errors. Is it affecting some of the other code in `parse_san`?

niklasf:
Hi, thanks for the pull request.

Regarding the issue on the previous pull request: It's always possible to get repositories back in sync without deleting your fork, but one good way to avoid the issue in the first place is working with feature branches. There must be many tutorials about that. Here's one: https://gist.github.com/vlandham/3b2b79c40bc7353ae95a

Regarding this patch: Using O as in upper-case letter o is definitely the correct notation according to the [specification](http://www.saremba.de/chessgml/standards/pgn/pgn-complete.htm#AEN237). But the mistake is common, so yeah ... why not be generous with what we accept as input.

Regarding the test failure: You can also run the test suite locally: `./test.py`. It will show that `Z0` now fails to parse down the line. How about just adding variants with zeros to the lists below?

Btw. the `chess/pgn.py` module already does a similar replacement.

https://github.com/niklasf/python-chess/blob/e4282b78f12be4a2019c3e9ac5b4646d7cc968e1/chess/pgn.py#L943-L948

This can be removed, then. Feel free to add that to the pull request.

tanmay2004:
I've made the edits, and thanks a lot for the tutorial and specification link! Also, I found [this](https://www.chess.com/forum/view/general/castling-notation), it was quite fun to look at people argue about the castling notation :)

PedanticHacker:
@tanmay, I think Travis is not happy because a testcase is missing for the code that you've written. Can you write a testcase for this as well?

niklasf:
This is already covered. The failing test was was due to `Z0` as explained above.

Issue #652, opened 2020-09-26
A new idea!


PedanticHacker:
What if we associate a move (a `chess.Move` object) and a board state in a Python dictionary, where the move would be the dictionary's key and the board state would be the dictionary's value?

Currently, we have a move stack, a simple Python list that stores moves (`chess.Move` objects). But alongside that (or perhaps even supersedingly), we could store "move state" objects (a Python dictionary of `chess.Move` objects where each such object stores a board state).

What do you think about this idea?

PR #653, opened 2020-09-26
Implement a new move container: move_state

PedanticHacker:
move_state is a Python dictionary that holds a move (a chess.Move object) as its key, and a board state (a chess._BoardState object) as its value. This PR references issue #652.


niklasf:
This cannot work, because the same move (as in `(from_square, to_square)`) can be played multiple times in the same game.

PR #670, opened 2020-10-06
Introduce restore_position() method in chess.Board()

PedanticHacker:
Introduced a new method in chess.Board() named restore_position() that restores the chessboard to a position by passing it an integer value as the index argument.

### DIFF
--- a/chess/__init__.py
+++ b/chess/__init__.py
@@ -1474,6 +1474,17 @@ class Board(BaseBoard):
         else:
             return self.copy(stack=False)
 
+    def restore_position(self: BoardT, index: int) -> BoardT:
+        """
+        Restores the board to a position by the given *index*.
+        
+        :raises: :exc:`IndexError` if the index is out of range.
+        """
+        if self._stack:
+            board = type(self)(None, chess960=self.chess960)
+            self._stack[index].restore(board)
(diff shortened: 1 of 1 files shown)


PedanticHacker:
@niklasf, is my implementation correct?

niklasf:
There's still a test or type checking failure, but it's definitely better than your previous approach. Mostly because the result no longer has an inconsistent history.

But that's because it does not have any history. Was that really your intention?

I think what you said in https://github.com/PedanticHacker/ChessFighter/issues/14#issuecomment-705752307 is a better idea. Just use `copy()`. You don't need `restore()` when you can just assign a different object to `board_state`.

PedanticHacker:
@niklasf, why does the `copy()` method of `chess.Board()` not include the following?

```
self.turn = board.turn
self.castling_rights = board.castling_rights
self.ep_square = board.ep_square
self.halfmove_clock = board.halfmove_clock
self.fullmove_number = board.fullmove_number
```

niklasf:
https://github.com/niklasf/python-chess/blob/ec90b9b7cfc8ed3b4aa08824e86c643beaa723cb/chess/__init__.py#L3498-L3502

PR #685, opened 2020-10-30
Improve a depreciation warning

PedanticHacker:
Improved the depreciation warning about the use of the *flipped* toggle in chess.svg.board().


niklasf:
Let's not make it more verbose.

PedanticHacker:
Maybe mention that the `flipped` parameter is going to be removed in chess v2.0.0?

PR #687, opened 2020-11-01
Add an 'a'

PedanticHacker:
Added an 'a' in "Fixed regression" to correctly say "Fixed a regression" now. Sorry for not noticing this in the previous PR.


PedanticHacker:
Why not merging this? Is the wording not correct if an 'a' is there?

PR #694, opened 2020-11-13
document `legal_drop_squares` and remove unnecessary copies in `_Craz…

kraktus:
…yhouseBoardState`

`legal_drop_squares` was not documented yet not used in the code source and considering it returns a `chess.SquareSet` I think it was intented to be exposed to the users.

when restoring a previous position form a `_CrazyhouseBoardState` it's not necessary to make a copy of the pocket since the instance is discarded. Also it's not intented to be publicly used. 

My only worry is that it can be considered as a breaking change stricto sensu. In that case I can make a commit only with doc change if deemed reasonable.

All test passed.

As always, thanks for making such a great lib!

### DIFF
--- a/chess/variant.py
+++ b/chess/variant.py
@@ -616,8 +616,8 @@ class _CrazyhouseBoardState(Generic[CrazyhouseBoardT], chess._BoardState[Crazyho
 
     def restore(self, board: CrazyhouseBoardT) -> None:
         super().restore(board)
-        board.pockets[chess.WHITE] = self.pockets_w.copy()
-        board.pockets[chess.BLACK] = self.pockets_b.copy()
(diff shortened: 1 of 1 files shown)


chess/variant.py
```diff
@@ -727,6 +727,13 @@ def legal_drop_squares_mask(self) -> chess.Bitboard:
             return chess.BB_EMPTY
 
     def legal_drop_squares(self) -> chess.SquareSet:
+        """
+        Get the squares where a legal drop is possible for the side to move.
```
PedanticHacker, line 731: It should say "Gets the squares..." and not "Get the squares...". It's a description, so it must be worded like that.

PedanticHacker (changes requested):

niklasf:
Thanks!

PR #710, opened 2020-12-13
draw knights moves using right-angled arrows

tewhalen:
Because I am a beginner at chess, it's easier for me to understand what's going on when arrows depicting a knight's move are drawn using right angles. I whipped up a patch that does this (and because I am not very good at math, I off-loaded some of the calculations onto the svg renderer). See attached image.
Screen Shot 2020-12-13 at 4 38 55 PM

### DIFF
--- a/chess/svg.py
+++ b/chess/svg.py
@@ -539,6 +539,9 @@ def __render_arrow(arrow, svg, colors, margin, orientation):
                 tail_rank - head_rank,
             )
         )
+        if not orientation:
+            # flip this angle if the board is flipped
+            angle = angle - 180
         svg_transform = "rotate({},{},{})".format(angle, xtail, ytail)
 
         lwo = 0.1 * SQUARE_SIZE  # line width offset
--- a/chess/svg.py
+++ b/chess/svg.py
@@ -120,7 +120,9 @@ class Arrow:
         if self.tail == self.head:
             return f"{color}{chess.SQUARE_NAMES[self.tail]}"
         else:
-            return f"{color}{chess.SQUARE_NAMES[self.tail]}{chess.SQUARE_NAMES[self.head]}"
+            return (
+                f"{color}{chess.SQUARE_NAMES[self.tail]}{chess.SQUARE_NAMES[self.head]}"
+            )
 
     def __str__(self) -> str:
         return self.pgn()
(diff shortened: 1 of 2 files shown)


niklasf:
Thanks for the pull request. The rendered result looks pretty good. A couple of comments:

* Please don't make changes to the coding style of the surrounding lines. It makes it hard to spot the actual changes. Instead, adjust your own additions to the surrounding coding style.
* I like the right-angled style, but I don't think it should be default. Arrows by `(2, 1)` are not guaranteed to be knight moves, semantically. Instead, `shape` or `style` could be a property of `Arrow`, similar to `color`.
* The transforms appear to be SVG Tiny compatible. We should double check that, though.

sshivaji:
Wanted to point out that the bishop move is incorrect, it's not diagonal in
the diagram. Am sure this is probably minor.

tewhalen:
> Wanted to point out that the bishop move is incorrect, it's not diagonal in the diagram. Am sure this is probably minor.

Yeah, that's a D5 -> G3 arrow, not intended to indicate a bishop's move.

>     * Please don't make changes to the coding style of the surrounding lines. It makes it hard to spot the actual changes. Instead, adjust your own additions to the surrounding coding style.

Sorry about that – the perils of relying on `black` for formatting help.

>     * I like the right-angled style, but I don't think it should be default. Arrows by `(2, 1)` are not guaranteed to be knight moves, semantically. Instead, `shape` or `style` could be a property of `Arrow`, similar to `color`.

That makes sense. I'll see if I can revise to make the arrow style selectable. I'm still learning the ins and outs of SVG, so I made the whole arrow (line + head) a polygon, which perhaps is not the right move, since it will be harder to style. I'll see about switching the lines back to lines or polylines and making the arrow head a marker.

>     * The transforms appear to be SVG Tiny compatible. We should double check that, though.

Hrrrm, I wasn't aware that was the target - I'll see if I can find a validator and check the output.

niklasf:
Sounds good. I don't mind the line/polygon decision. Whichever is easier to implement.

PedanticHacker:
Although the idea of this PR is very interesting, the implementation of it is very bad.

Here's my rant about it:
`(1)` Unnecessary code style changes, as @niklasf 
already pointed out (and a very bad style proposed, if you ask me); `(2)` no type annotations (that's why there were errors); `(3)` neglecting PEP 8 completely, especially with comments (a comment should start start with a capital letter, and end with a dot); `(4)` you don't gain anything by prepending two underscores to a function (one underscore is enough).

I propose this PR is deleted, and a new attempt is made.

tewhalen:
> Although the idea of this PR is very interesting, the implementation of it is very bad.

I'll poke around with it a little more. I learned Python before PEP 8 was written and I haven't programmed professionally in over 15 years, so my style is pretty bad by modern standards.

PedanticHacker:
I noticed that Fritz chess GUI does not have L-shaped knight move arrows. So I think it's fine not having that. I guess there's no need for this.

PR #716, opened 2020-12-19
Faster Piece.__hash__ function

bartekspitza:
Avoiding the multiplication yields slightly better performance (I am writing an engine, don't mean to be nitpicky :D )


niklasf:
Thanks. You might find that many of the higher level functions are even less designed with chess engine performance in mind, but patches to improve it are always welcome.

PR #720, opened 2020-12-24
Fix wrong declaration in Options section

SzymonStasiowski:
The snippet is not working because in following lines "engine" is called instead of declared "protocol"


niklasf:
Thanks!

PR #729, opened 2021-01-06
Fix variation_san() output format + its docstring

PedanticHacker:
The sequence of chess moves of a variation are written as, e.g., "1.e4 e5 2.Nf3 Nc6", so there's no space after the dot. In all chess publications I could find, I found this format.


niklasf:
Matter of taste. We'll go with the [PGN standard](http://www.saremba.de/chessgml/standards/pgn/pgn-complete.htm#c2.3).

PR #735, opened 2021-02-13
Implement Xboard usermove feature

MarkZH:
At the request of the Xboard engine with "feature usermove=1", incoming move are prefixed with the keyword "usermove".


niklasf:
Thanks! Good change.

PR #747, opened 2021-03-08
Add animation of the last move.

themactep:
I have rearranged final SVG by layers: first, the whole board, then pieces, so that board squares won't overlap the animated piece.

### DIFF
--- a/chess/svg.py
+++ b/chess/svg.py
@@ -384,13 +384,13 @@ def board(board: Optional[chess.BaseBoard] = None, *,
 
                 # Render animated move.
                 if animation:
-                    if square == lastmove.to_square:
-                        dx = int((math.fmod(lastmove.from_square, 8) - math.fmod(lastmove.to_square, 8)) * SQUARE_SIZE)
-                        dy = (math.ceil(lastmove.to_square / 8) - math.ceil(lastmove.from_square / 8)) * SQUARE_SIZE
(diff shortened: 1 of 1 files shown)


niklasf:
Thanks for sharing.

I am getting this error when trying this:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-7-c023e3573820> in <module>
----> 1 chess.svg.board(chess.Board(), lastmove=chess.Move.from_uci("f3g1"), animation=True)

~/Projekte/python-chess/chess/svg.py in board(board, orientation, lastmove, animation, check, arrows, squares, size, coordinates, colors, flipped, style)
    391                             "dur": f"1s",
    392                             "repeatCount": f"1",
--> 393                             "path": f"M{lmdx:d},{lmdy:d} 0,0",
    394                         })
    395 

ValueError: Unknown format code 'd' for object of type 'float'
```

I expect there might also be problems with promotions, castling, en passant captures and chess variants (like Atomic explosions and maybe Crazyhouse drops). I am merging regardless and partially reverting again, just to get the reordering that this patch includes.

themactep:
Right. This might require some further ammendment as I'm not a python guy, I mostly use ruby. I copied that {:d} notation from the block you use for rendering pieces. Maybe casting lmdx, lmdy to integers could help.

PR #752, opened 2021-03-12
Send Xboard level command in sudden death games

### LINKED ISSUE #742: Xboard time control implemented incorrectly
I believe that the implementation of Xboard time control is incorrect in two aspects:

1. The `level` and `st` commands are sent on every move, when they should only be sent before a game starts and only resent when time control changes. Only `time` and `otim` should be sent before most moves.

From the [Xboard spec](https://www.gnu.org/software/xboard/engine-intf.html#11):

> Example: if at the start of the game "level 40 60 0" was given (40 moves per hour), and the engine receives "level 20 22 0" just before move 41, it should understand that it should do the next 20 moves in 22 minutes (pehaps because the secondary session was 20 moves per 15 minutes, and it had 7 minutes left on its clock after the first 40 moves). 

2. If a time control is used with no `remaining_moves` and no `increment` (e.g., each player has 10 minutes for all their moves), then the `level` command is never sent, when `level 0 10 0` should be. This is because, as it is now, `level` is only sent when `remaining_moves` or `increment` are truthy, which zero is not ([line 2050 in chess/engine.py](https://github.com/niklasf/python-chess/blob/1b21b80fb9b574085c245e1261b103cfa5cdcf6f/chess/engine.py#L2050)). Since `st` and `level` time controls are mutually exclusive, perhaps something like the following could work:
```python
if limit.time is not None:
    engine.send_line(f"st {max(0.01, limit.time)}")
else:
    base_mins, base_secs = divmod(int((limit.white_clock if board.turn else limit.black_clock) or 0), 60)
    engine.send_line(f"level {limit.remaining_moves or 0} {base_mins}:{base_secs:02d} {increment}")
```

For reference, [here's a link to how I'm avoiding these issues with the upgraded version of python-chess for lichess-bot](https://github.com/MarkZH/lichess-bot/blob/01e2a36f9929afe6c2f8418ff162911d66fd4eaf/engine_wrapper.py#L135). The `level` command is sent manually at the beginning of a game using the `SimpleEngine.protocol.send_line()` command (line 132). Thereafter, I create a `Limit` instance with only the `white_clock` and `black_clock` fields set so only `time` and `otim` commands are sent to the engine (line 139). If I'm doing this wrong, please advise on the correct way to do this.

MarkZH:
Games with a base time limit but no moves to reset
(Limit.remaining_moves) and no increment still require a level command.
Since st and level are mutually exclusive, use the presence or absence
of time-per-move (Limit.time) to decide which time control command to
send.

Fixes #742


niklasf:
Thanks!

PR #755, opened 2021-03-13
make sure that other pieces don't overlap animation

themactep:
If `lastmove` is present, the moved piece won't be rendered along with other pieces (skipped), and rendered separately, so it does not go under other later rendered pieces while moving, if `animation` is requested.

Here's the resulting SVG: https://php.themactep.com/chess-test.svg

### DIFF
--- a/chess/svg.py
+++ b/chess/svg.py
@@ -233,6 +233,7 @@ def piece(piece: chess.Piece, size: Optional[int] = None) -> str:
 def board(board: Optional[chess.BaseBoard] = None, *,
           orientation: Color = chess.WHITE,
           lastmove: Optional[chess.Move] = None,
+          animation: bool = False,
           check: Optional[Square] = None,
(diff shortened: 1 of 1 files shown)


niklasf:
Thanks! Looks like this fixes the exception from #747. The concerns around castling, en passant, and variants still apply.

themactep:
Nicklas, is it possible to make Move to return details of a complex move?
Say, if the move is a castling then it also returns rook's move? Something like 
```
lastmove = board.peek()
if (lastmove.castling == True):
    print(lastmove.rook.from_square)
    print(lastmove.rook.to_square)
```
I think I could make it to generate both pieces animated. Just need to get the other party form the move data per se.

niklasf:
Sorry to reject it now, after you have put in more work. Considering that implementing this cleanly will require breaking changes (like looking at the board history, when all we have is a `BaseBoard`), I think it's best not to add animations at this point.

Issue #757, opened 2021-03-16
UCI ponderhit not implemented


MarkZH:
When the method `UciProtocol.play()` is called with `ponder=True`, python chess sets the engine to pondering with a `go ponder ...` command if the engine responded with `bestmove ... ponder ...`. However, upon the next call to `play()`, even if the ponder move was played, the pondering is stopped and a new search is started. What should happen is that, if the ponder move is the last move in `board.move_stack`, `ponderhit` is sent to the engine and python-chess waits for a `bestmove` response from the engine. As it is now, engines will throw away their pondering results when they receive `stop` and start a new move search, negating the pondering work.

Or is it that I'm wrong and `play()` should not be called on a pondering engine? Is it possible to manually send `engine.protocol.send_line("ponderhit")` and get the pondering result some other way?

Secondary but related issue: Just before the `go ponder ...` command is sent to the engine, `engine._position()` is called ([here](https://github.com/niklasf/python-chess/blob/c5739f13b64708e371e2be29a950b14096267a5b/chess/engine.py#L1586)) with the `bestmove` and `ponder` pushed onto the board. However, this results in a truncated move list because the `engine.board` is used to create the pondering board. But, this board is created with `board.copy(stack=False)`, so the pondering board does not have the entire game history, which may lead to inaccurate evaluations due to not seeing three-move repetition draws. Using `ponder_board = board.copy()` instead of `engine.board` here would be better.

I can create a pull request for the second issue, but I'm less sure of how to implement the `ponderhit` functionality.

niklasf:
Mhh, yes, this seems broken by design, and I don't know how to fix it.

Ideally, the next call to `play()` would take care of everything, like continuing with `ponderhit` or setting up a new position. But that's not possible, because `play()` will commonly get new search limits, but `ponderhit` has no way to apply them.

For example, I expect the most common use case for pondering to be playing timed matches. So the engine moves, and we set it to ponder. `go ponder wtime <our remaining time> btime <???>`, where `<???>` would be known only after our opponent's move, the moment we would send `ponderhit`, but then it's too late.

It is however not broken as badly as one may think, at least with the most popular engines. Thanks to the transposition table, restarting search on the same position is almost as good as continuing.

MarkZH:
I've created a pull request with what I think is a correct implementation of `ponderhit`. I'll correct it as needed.

[...]

PR #758, opened 2021-03-17
Use full move stack for UCI pondering

MarkZH:
Refer to secondary part of issue #757. Before sending a `go ponder` command to a UCI engine, this will send the entire move stack to the engine in a `position` command instead of just the current FEN followed by the `bestmove` and `ponder` move.


niklasf:
Thanks! Good find and solution.

PR #779, opened 2021-05-04
Implementation of HordeBoard.has_insufficient_material

stevepapazis:
Hey, there

This pull request implements the insufficient material method for Horde.

First we need to determine the horde checkmating patterns that use the least amount of material. 
That is fairly easy to do in positions where white has queens, rooks and pawns or when white has more than three pieces; for example a queen and any other white piece mate the black king, pawns promote to queens or knights, etc.

Positions with minor pieces are a bit trickier, since they're more of them and a bishop and a knight can deliver a double check. But even those can be worked out by hand. Then we can use a brute force approach to validate that we haven't missed out any of them, similarly to how endgame tablebases are generated.  

We quantify those patterns into conditions that can be used to check any given position against at run time. 

[Here](https://github.com/stevepapazis/horde-insufficient-material-tests) is a little more information on horde checkmating patterns and generation of test positions.

### DIFF
--- a/chess/variant.py
+++ b/chess/variant.py
@@ -457,9 +457,206 @@ class HordeBoard(chess.Board):
         return bool(self.occupied) and not self.occupied_co[not self.turn]
 
     def has_insufficient_material(self, color: chess.Color) -> bool:
-        # TODO: Could detect some cases where the Horde can no longer mate.
-        return False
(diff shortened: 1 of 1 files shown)


niklasf:
Oh, awesome, thanks! It will take me some time to review this.

PR #783, opened 2021-05-16
Notify engines of draw offers

### LINKED ISSUE #781: Engine communication outside play() and other methods
I'm trying to implement the handling of draw offers and resignations in [lichess-bot](https://github.com/ShailChoksi/lichess-bot/). I have a question about getting messages sent by an engine when not invoking `play()`, `analyze()` or other engine methods.

Here's a stripped down version of the main game-playing loop with an XBoard engine:

```python
engine = chess.engine.SimpleEngine.popen_xboard(configuration)
while True:
    game = get_game_state_from_server()
    if game.is_over():
        break
    play_result = engine.play(game.board, game.clock)
    send_move_to_server(play_result.move)
```

In order to handle draw offers, I've changed the loop to the following:

```python
engine = chess.engine.SimpleEngine.popen_xboard(configuration)
while True:
    game = get_game_from_server()
    if game.is_over():
        break
    if game.draw_offered():
        engine.protocol.send_line("draw")
    play_result = engine.play(game.board, game.clock)
    send_move_to_server(play_result.move, play_result.draw_offered)
```

My question is about the time between the `send_line("draw")` and the `engine.play()`. If the engine sends "offer draw" (to accept the draw) before the `engine.play()` method is invoked, is there a way to retrieve that communication? Or, is there a better way to go about this?

MarkZH:
Only works for XBoard engines since the UCI protocol has no facility for
offering or accepting draws.

Closes #781

### DIFF
--- a/chess/engine.py
+++ b/chess/engine.py
@@ -1078,7 +1078,7 @@ class Protocol(asyncio.SubprocessProtocol, metaclass=abc.ABCMeta):
         """
 
     @abc.abstractmethod
-    async def play(self, board: chess.Board, limit: Limit, *, game: object = None, info: Info = INFO_NONE, ponder: bool = False, root_moves: Optional[Iterable[chess.Move]] = None, options: ConfigMapping = {}) -> PlayResult:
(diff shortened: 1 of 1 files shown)


niklasf:
Thanks!

PR #792, opened 2021-06-21
Correcting likely copy / paste error in test

chrkorn:
I think the caro kann mate is suppsoed to be tested here, not the starting position


niklasf:
Thanks!

PR #793, opened 2021-06-28
piece_map_by_color added

JJBT:
Very useful


niklasf:
Hi, thanks! Maybe we could have `def piece_map(self, *, mask: Bitboard = BB_ALL)` instead of this, or as a way to implement it without duplication?

PR #795, opened 2021-07-08
Hz

HamzhSuilik:
I added a advantage method to Board class to determine which player has the highest chance of winning , also I added additional documentation  file


niklasf:
Sorry, but evaluation functions are out of scope for this library. They are too subjective. See https://python-chess.readthedocs.io/en/latest/core.html for more complete automatically generated documentation.

PR #799, opened 2021-07-24
Fix bratko_kopec movetime description

analog-hors:
The description for the `movetime` parameter in the `bratko_kopec` example is clearly incorrect.


niklasf:
Thanks! Good catch. Looks like I missed it when updating the example to python-chess 1.0 (https://github.com/niklasf/python-chess/commit/62526edcbf1a6ce40d7e75c1c4074bdf052fef18#diff-44d352b2ba76e08c992a0705668ed06f82e188db1572197a2282fd0cfcfc9785L32)

PR #816, opened 2021-11-01
Convert eleven statements to the usage of augmented assignments

elfring:
[Augmented assignment statements became available with Python 2](https://docs.python.org/3/whatsnew/2.0.html#augmented-assignment "Augmented assignments").
:thought_balloon: Thus improve eleven source code places accordingly.


niklasf:
Thanks!

Issue #836, opened 2021-12-17
Elapsed Move Time in the pgn file


aipv:
Hello guys, great library!

I am working on the pgn files generated from electrical boards, there are ``[%emt ...]`` annotations (Elapsed Move Time) in the pgn files, they are quite useful as they show the elapsed time that the players used for the commented move. However, currently it cannot be read from the GameNode directly.

Do you think it makes sense to introduce emt() and set_emt() functions in the GameNode? similar like the clock() function and set_clock() function for the ``[%clk ...]`` annotations?  Thank you very much!

niklasf:
Yes, indeed, this would be nice to have.

niklasf:
Added via #837.

niklasf:
Included in v1.8.0.

PR #837, opened 2021-12-17
support of emt (Elapsed Move Time) command in pgn comments

aipv:
Elapsed Move Time in the pgn file #836. A simple pull request follows the implementation of clock() and set_clock() function, please feel free the modify it if it can be better. thank you.

### DIFF
--- a/chess/pgn.py
+++ b/chess/pgn.py
@@ -121,6 +121,7 @@ SKIP_MOVETEXT_REGEX = re.compile(r""";|\{|\}""")
 
 
 CLOCK_REGEX = re.compile(r"""\[%clk\s(\d+):(\d+):(\d+(?:\.\d*)?)\]""")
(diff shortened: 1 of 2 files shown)


niklasf:
Thanks! Will also need some tests in `test.py`.

aipv:
Thank you very much for the comments and sorry for missing the test codes. The pull request is updated to include the "emt" comment in test_annotations(), and an individual test case in test_float_emt(). When set_emt() to a negative value, it sets the Elapsed Move Time to 0:00:00, similar as the set_clock(), I think it makes sensor and hope it is ok.  Please let me know if further test cases are required, thank you very much!

niklasf:
Perfect. If you or anyone else wants to use this before the next release
```sh
pip install git+https://github.com/niklasf/python-chess.git
```
would work.

aipv:
Yes, it works. thank you very much!

PR #840, opened 2021-12-23
Update __init__.py

rtk-rnjn:
`to_list()` method in LegalMoveGenerator


niklasf:
Thanks, but I would expect a method like that to give a `List[chess.Move]`, which is already available via `list(board.legal_moves)`. As for getting a list of moves in standard algebraic notation, I believe your implementation is concise enough, so I don't think we need to provide this shortcut.

PR #850, opened 2022-01-28
Add ascii board as description of svg

kdave:
The tag <desc> in SVG can store a textual description, eg. for better
accessibility. For the chess board there could be the string
representation (basically str(chess.Board()) and with <br/> this is
readable in text browsers like links or w3m when the svg is embedded in
html.


niklasf:
Thanks! Would it possible (and make sense) to use `<pre>...</pre>`?

kdave:
Ah yeah that's better. Updated and pushed.

PR #868, opened 2022-04-03
Syzygy probing: Use list instead of dict

kraktus:
Tests passed. Looking at the code those variables where only accessed by index so I'm not sure why they were defined as dict. Could be done for visibility/backward compatibility, if then feel free to close the PR.

`PawnFileData` could probably be further refractored not to use dicts at all anymore, since it looks like the rust implementation only use `ArrayVec`.


niklasf:
Thanks. I agree that dicts make no sense, there.

PR #890, opened 2022-06-06
Mention Chess Match Viewer (PyScript) in README

nmstoker:
Would it be okay to mention this PyScript demo here? Hope the formatting is suitable.


niklasf:
Thanks for sharing, that looks pretty good! I didn't know PyScript is a thing. Can you please have a look at the build error? Probably just a blank line missing.

nmstoker:
Yes, it was my mistake not noticing I'd inadvertently removed a final blank line after the list. Added that blank line back, so now the only change is the addition of the bullet point i meant to add.

Btw big thank you for your amazing chess module - it made producing the PyScript tool remarkably straightforward (was able to leverage the Jupyter / SVG code)

Issue #896, opened 2022-06-16
Add fill for black pawn SVG so it renders correctly on dark backgrounds


jmhammond:
Here's the problem. In emacs, viewing the black pawn SVG looks like this with a black background:
<img width="470" alt="image" src="https://user-images.githubusercontent.com/84139/173984087-3839c0c0-1caa-4e16-aed1-8975512beb22.png">

If I change my background to light, it looks like this:

<img width="514" alt="image" src="https://user-images.githubusercontent.com/84139/173984154-e9cc7e94-ea70-4c72-8dc8-81621a60c8c7.png">

What needs to happen to fix it is to edit line https://github.com/niklasf/python-chess/blob/193a3ba0994cb218be31499af2a1786bd967e645/chess/svg.py#L39 by adding `fill="#000"`  This fixes the issue and is consistent with the design of the white pawn which specifies `fill="#fff"`.

niklasf:
Thanks!

niklasf:
Thanks!

niklasf:
via #897

PR #897, opened 2022-06-16
Fill the black pawn with black color

jmhammond:
This addresses #896 when rendering the board on a dark background.


niklasf:
Thanks!

Issue #900, opened 2022-06-22
PGN: Moves are not parsed if a header name contains a colon


laniakea64:
I have a PGN file with a header `[BlackType: "program"]`.  If I try to load this file using `chess.pgn`, none of the moves are loaded, it only loads the headers.

Reduced testcase:
`test.pgn`
```
[Event "Test"]
[Date "2022-06-22"]
[Result "0-1"]
[BlackType: "program"]

1. f4 e5 2. g4 Qh4
0-1
```
```
import chess.pgn
with open('test.pgn', 'r') as f:
  game = chess.pgn.read_game(f)

print(game, game.end().board(), sep='\n')
```

The following monkeypatch based on the [PGN spec](http://www.saremba.de/chessgml/standards/pgn/pgn-complete.htm#c7) gets it working as expected -
```
chess.pgn.TAG_REGEX = re.compile(r"^\[([A-Za-z0-9_+#=:-]+)\s+\"([^\r]*)\"\]\s*$")
chess.pgn.TAG_NAME_REGEX = re.compile(r"^[A-Za-z0-9_+#=:-]+\Z")
```

Would it be helpful if I make a pull request of that as a proper patch?

niklasf:
Yes, please.

PR #901, opened 2022-06-22
pgn: Support all valid symbol continuation characters in tag names

laniakea64:
Fixes https://github.com/niklasf/python-chess/issues/900


niklasf:
Thanks!

Issue #904, opened 2022-07-07
Invalid PGN header ends header section


niklasf:
[...]  While writing code that uses `chess.pgn`'s handling of multiple PGN games from one file, I found that although https://github.com/niklasf/python-chess/pull/901 fixes some specific cases of this issue, there is an underlying root issue it doesn't address:

The issue is not that moves aren't parsed.  The root issue here is that if `chess.pgn` hits an invalid tag, it
 - determines that point in the file to be the end of the game;
 - drops not just the bad tag, but all remaining tags as well;
 - treats the moves as a completely separate game with no tags except Result set.

This seems not to match the [documented behavior](https://python-chess.readthedocs.io/en/latest/pgn.html#chess.pgn.read_game) -

> The parser is relatively forgiving when it comes to errors. It skips over tokens it can not parse. By default, any exceptions are logged and collected in `Game.errors`.

Based on that quote, it seems the expected behavior would be for `chess.pgn` to parse the file as if the bad tag weren't there; or, failing that, log something to `.errors` of the first `chess.pgn.Game` parsed from the malformed game.  Or am I missing something?

Test code and test invalid PGN file:
```
import chess.pgn

with open('invalid_multi.pgn', 'r') as f:
  gamelist=[]
  while True:
    gamelist.append(chess.pgn.read_game(f))
    if gamelist[-1] is None: break

print(gamelist)
for game in gamelist:
  print('\n---------------------\n')
  print(game)
  if game:
    print(game.errors)
    print(game.end().board())
```

`invalid_multi.pgn`:
```
[Event "Test"]
[Date "2022.06.22"]
[Result "0-1"]
[BlackType: "program"]

1. f4 e5 2. g4 Qh4
0-1

[Event "Test2"]
[Date "2022.07.05"]
[Result "1-0"]
[GarbageTag!!!!! "skipme"]
[White "test1"]
[Black "test2"]

1. e4 e5 2. Qh5 Ke7 3. Qxe5#
1-0
```

_Originally posted by @laniakea64 in https://github.com/niklasf/python-chess/issues/900#issuecomment-1175528570_

niklasf:
Via #906.

PR #906, opened 2022-07-09
pgn: Skip over invalid or malformed headers

laniakea64:
From https://github.com/niklasf/python-chess/issues/904#issue-1297715055 -
> Based on that quote, it seems the expected behavior would be for `chess.pgn` to parse the file as if the bad tag weren't there; 

With this patch `chess.pgn` now does this in most cases.


niklasf:
Thanks! Looks sensible.

PR #909, opened 2022-07-21
Adding flipped option to unicode representation

DonLarry:
chess.svg.board has the option to be flipped, so I thought Unicode representation could also have a reversed/flipped option, which is the feature this commits adds.

### DIFF
--- a/chess/__init__.py
+++ b/chess/__init__.py
@@ -1221,7 +1221,7 @@ class BaseBoard:
 
         return "".join(builder)
 
-    def unicode(self, *, invert_color: bool = False, borders: bool = False, empty_square: str = "⭘", flipped:bool = False) -> str:
+    def unicode(self, *, invert_color: bool = False, borders: bool = False, empty_square: str = "⭘", orientation:Color = WHITE) -> str:
         """
         Returns a string representation of the board with Unicode pieces.
         Useful for pretty-printing to a terminal.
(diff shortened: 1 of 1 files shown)


chess/__init__.py
```diff
@@ -1221,14 +1221,19 @@ def __str__(self) -> str:
 
         return "".join(builder)
 
-    def unicode(self, *, invert_color: bool = False, borders: bool = False, empty_square: str = "⭘") -> str:
+    def unicode(self, *, invert_color: bool = False, borders: bool = False, empty_square: str = "⭘", flipped:bool = False) -> str:
```
niklasf, line 1224: `orientation` is preferable in `chess.svg`. Let's use it here, too.

https://github.com/niklasf/python-chess/blob/aa0a04f37da9339b7d3a983aecbe9c97935fc0d3/chess/svg.py#L292-L293

chess/__init__.py
```diff
@@ -1221,14 +1221,19 @@ def __str__(self) -> str:
...
         Useful for pretty-printing to a terminal.
 
         :param invert_color: Invert color of the Unicode pieces.
         :param borders: Show borders and a coordinate margin.
         """
+        board = self.copy()
+        if flipped:
+          board.apply_transform(flip_vertical)
+          board.apply_transform(flip_horizontal)
+
```
niklasf, line 1236: This transformation will not change the coordinates if `board.unicode(..., borders=True)`. Best to change the actual iteration order instead.

niklasf (changes requested):

DonLarry:
Hi @niklasf, I think I've applied all the suggestions you gave me :)

niklasf:
Thanks!

PR #911, opened 2022-07-27
Update docs

johndoknjas:
I was trying to open a pgn, but received a UnicodeDecodeError. Adding 'r' and errors='replace' as arguments to open() solved the issue for me. 
Used the top answer here: https://stackoverflow.com/questions/35028683/python3-unicodedecodeerror-with-readlines-method

This PR just updates the documentation to include this, in case anyone else has a similar problem.


niklasf:
Thanks. I hesitate to add this advice, because the proper fix would be to specify the correct encoding.

> It is the caller’s responsibility to open the file with the correct encoding. PGN files are usually ASCII or UTF-8 encoded, [...]

Will qualify it a bit ...

PR #912, opened 2022-07-28
Add draw() method to tell engines of draw offers

MarkZH:
As of now, an Xboard engine can only told about a draw offer when its
opponent makes a move and the play() method is called. By adding this
method, an engine can be told of a draw offer during thinking, pondering,
or any other time.

### DIFF
--- a/chess/engine.py
+++ b/chess/engine.py
@@ -1095,6 +1095,12 @@ class Protocol(asyncio.SubprocessProtocol, metaclass=abc.ABCMeta):
         """
 
     @abc.abstractmethod
+    def draw(self) -> None:
+        """
+        Tells engine that its opponent has offered a draw.
+        """
+
+    @abc.abstractmethod
     async def configure(self, options: ConfigMapping) -> None:
         """
         Configures global engine options.
@@ -1445,6 +1451,9 @@ class UciProtocol(Protocol):
                     LOGGER.warning("%s: Unexpected engine output: %r", engine, line)
 
         return await self.communicate(UciPingCommand)
+
+    def draw(self) -> None:
+        pass
 
     def _changed_options(self, options: ConfigMapping) -> bool:
         return any(value is None or value != self.config.get(name) for name, value in _chain_config(options, self.target_config))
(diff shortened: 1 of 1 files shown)


niklasf:
This API has some problems. How can the engine accept? Is the offer still valid when sending `draw()` followed by `play()`?

chess/engine.py
```diff
@@ -2772,6 +2785,9 @@ def ping(self) -> None:
             future = asyncio.run_coroutine_threadsafe(coro, self.protocol.loop)
         return future.result()
 
+    def draw(self) -> None:
+        self.protocol.draw()
```
niklasf, line 2789: Simple function, but still needs to check that the engine is alive and be called from the event loop, much like https://github.com/niklasf/python-chess/blob/2379c55bcadd6ef9f2acb2e2f7f2476094be5ec1/chess/engine.py#L2892-L2894

MarkZH:
I've read some more of the documentation, and I think this PR should be closed. Read on for details.

The engine does not have to respond to the draw offer until it chooses a move, if ever. The engine cannot assume that the draw offer is still valid if it responds with `offer draw`, so it still has to finish thinking about a move. However, now that I've read more of the documentation, I think this arbitrary timing of the sending of a draw offer is unnecessary. From [here](http://hgm.nubati.net/CECP.html):

>draw
>
>The opponent offers you a draw. Respond with 'offer draw' if you want to accept it; just ignore it to reject the offer. Do not count on the offer still to be on the table when you accept it. Sending of this command can be suppressed by sending 'feature draw=0', which could be useful if your engine's thinking would be aborted by receiving a 'draw' command. **GUIs should protect you from this by relaying draw offers only just before you get the opponent's move, so that it can neither interfere with thinking or pondering, but not all GUIs do.**

Emphasis mine. I thought it would be useful to inform engines of draw offers immediately since lichess sends `gameState` updates with the draw offer immediately when offered. However, it will also include the draw offer on every move while the offer is valid. I think sending the offer can wait until it's the bot's turn to move.

PR #915, opened 2022-08-08
Add MiniMax, Alpha-Beta , NegaMax Algorithm

### LINKED ISSUE #361: Tree generation (MinMax, Alpha-Beta, Negascout, etc.)
Hello python-chess community, 

I am wondering why the current version and next one 1.0 do not consider adding move tree generation with classic algorithms: MinMax, Alpha-Beta, etc. as part of the API? Thanks in advance for the answer

mph

GarboChess:
Closes #361

### DIFF
--- a/examples/minimax.py
+++ b/examples/minimax.py
@@ -1,0 +1,122 @@
+import chess
+import random
+
+
+class Person:
+  def __init__(self, name, age):
+    self.name = name
+    self.age = age
+
+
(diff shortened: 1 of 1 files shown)


niklasf:
Hi, thanks for sharing. Let's add any missing pieces to https://github.com/Mk-Chan/python-chess-engine-extensions instead. It's a library and can serve as an example at the same time.

Issue #659, opened 2020-09-28
Ideas for breaking changes


niklasf:
Changes to consider when the opportunity arises.

* `chess.svg.board(..., flipped)` -> `chess.svg.board(..., orientation)`
* Remove `chess.pgn.BaseVisitor.parse_san()`

PedanticHacker:
(1) Niklas, why do you want to rename the `flipped` argument name of `chess.svg.board()` to `orientation`? 🤔  If you think of renaming it, you can do it right now. Make it as a 1.x breaking change. Everyone have not upgraded to 1.x yet, so the opportunity is basically right now. Version 1.0.2 can introduce this. The changelog should explain this and people will adapt. And there's only one simple name to change, not a hassle at all.

(2) Do you want to remove `parse_san()` abstract method of the `BaseVisitor()` abstract base class altogether, or just remove `return board.parse_san(san)` and make it `pass` instead?

PedanticHacker:
Here's an idea of a breaking change: since the introduction of passing a dictionary as the `colors` argument to `chess.svg.board()`, we can now remove the `style` argument in there. Thoughts?

niklasf:
Good point about `style`. And no, I will not sneak breaking changes into minor releases. This is just a list of things to consider in the future.

PedanticHacker:
A little off-topic here, but python-chess v1.1.0 is still not available via pip to upgrade, and the latest version for python-chess on PyPI shows as 1.0.1. Have you not uploaded the new version yet?

niklasf:
* Remove `chess.engine.PovWdl` bc for `Tuple[int, int, int]`.

PedanticHacker:
- Remove `flipped` and `style` parameters of `chess.svg.board()`.

niklasf:
* Distinguish PGN root and child nodes for typing

khink:
Not sure if i should post here, but here goes:

I'm looking at what is needed to create an interface/engine for a 5x5 board. Am i correct in assuming this would be a major, breaking change in this module? Looking at the [constants](chess/__init.py) like SQUARES, FILE_NAMES, RANK_NAMES, and also the various [variants](chess/variants.py), this is not easily changed in the current situation, right?

[...]

PR #921, opened 2022-09-23
Make Pieces immutable and return them from cache

akx:
This PR makes Piece a `frozen` dataclass and adds a cache for them. For end users (and the library itself of course) this change is invisible API-wise.

However, there is a visible performance increase (even if the test suite is not necessarily the best indicator):

```
hyperfine 'git checkout master; git clean -fdx; py.test --count=10 test.py' 'git checkout immutable-pieces; git clean -fdx; py.test --count=10 test.py'
```

```
Benchmark 1: git checkout master; git clean -fdx; py.test --count=10 test.py
  Time (mean ± σ):      6.405 s ±  0.411 s    [User: 5.249 s, System: 0.404 s]
  Range (min … max):    6.120 s …  7.537 s    10 runs

Benchmark 2: git checkout immutable-pieces; git clean -fdx; py.test --count=10 test.py
  Time (mean ± σ):      6.246 s ±  0.131 s    [User: 5.218 s, System: 0.394 s]
  Range (min … max):    6.126 s …  6.591 s    10 runs

Summary
  'git checkout immutable-pieces; git clean -fdx; py.test --count=10 test.py' ran
    1.03 ± 0.07 times faster than 'git checkout master; git clean -fdx; py.test --count=10 test.py'
```

Let me know what you think :)

### DIFF
--- a/chess/__init__.py
+++ b/chess/__init__.py
@@ -447,7 +447,7 @@ SAN_REGEX = re.compile(r"^([NBKRQ])?([a-h])?([1-8])?[\-x]?([a-h][1-8])(=?[nbrqkN
 FEN_CASTLING_REGEX = re.compile(r"^(?:-|[KQABCDEFGH]{0,2}[kqabcdefgh]{0,2})\Z")
 
 
-@dataclasses.dataclass
+@dataclasses.dataclass(frozen=True)
 class Piece:
     """A piece with type and color."""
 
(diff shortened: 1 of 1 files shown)


niklasf:
Hi, thanks! I like this patch.

Even as is, pieces are *intended* to be immutable (no setters, provides `__hash__`). But that's not documented, so this is technically still a bc break and needs to [wait for a future version 2.x](https://github.com/niklasf/python-chess/issues/659).

Phillyclause89:
> Hi, thanks! I like this patch.
> 
> Even as is, pieces are _intended_ to be immutable (no setters, provides `__hash__`). But that's not documented, so this is technically still a bc break and needs to [wait for a future version 2.x](https://github.com/niklasf/python-chess/issues/659).

Please don't ever change that lol!

https://github.com/Phillyclause89/ChessMoveHeatmap/blob/e482a42978f824ab477e4c0e2ee2467e004e5b71/heatmaps/__init__.py#L301C1-L314C49
```python
class ChessMoveHeatmapT(GradientHeatmap):
    """A gradient heatmap that additionally tracks chess piece counts per square.

    This class extends `GradientHeatmap` by maintaining, for each square, a dictionary
    mapping each chess piece to a floating point value. This value can represent an intensity
    or count of moves involving that piece on the square.

    Attributes
    ----------
    piece_counts : numpy.ndarray of dict
        A NumPy array of shape (64,) where each element is a dictionary mapping
        chess `Piece` objects to a `float` count.
    """
    _piece_counts: NDArray[Dict[Piece, float64]]
```

PR #925, opened 2022-10-12
Added chessmart: https://github.com/blindpandas/chessmart

mush42:
Added [Chessmart](https://github.com/blindpandas/chessmart) an accessible chessboard that allows blind chess players to play against Stockfish.


niklasf:
Cool! Thanks for sharing.

PR #927, opened 2022-10-14
Add specific move error exceptions

trevorbayless:
Currently, when an invalid, illegal, or ambiguous move is attempted a `ValueError` is raised (with differing messages regarding the error). If a library user is interested in displaying a custom error to the user such as: "_That move is ambiguous_" the only option in this case is to parse the `ValueError` message for "_ambiguous_" and then handle presenting a message regarding the ambiguous move accordingly.

This pull request brings in _specific_ move exceptions (`InvalidMoveError`, `IllegalMoveError`, `AmbiguousMoveError`, and `PromotionMoveError`) and raises them accordingly. All of these new exceptions inherit from `ValueError` and provide the same messages as before to allow for full backwards compatibility since the base `ValueError` can still be caught in all of these situations.

This is purely just a "this would be nice to have" pull request as I know there are existing unit tests in place to verify the messages are not changed of the currently raised `ValueErrors`, but it would be nice to not have to parse the messages.

I completely understand declining this pull request if you believe this opens Pandora's box with the other existing `ValueErrors` (like FEN parsing, etc) or if you think this would be difficult to maintain.

### DIFF
--- a/chess/__init__.py
+++ b/chess/__init__.py
@@ -165,6 +165,18 @@ class Outcome:
         return "1/2-1/2" if self.winner is None else ("1-0" if self.winner else "0-1")
 
 
+class InvalidMoveError(ValueError):
+    """Raised when the attempted move is invalid in the current position"""
+
+
+class IllegalMoveError(ValueError):
+    """Raised when the attempted move is illegal in the current position"""
+
+
+class AmbiguousMoveError(ValueError):
+    """Raised when the attempted move is ambiguous in the current position"""
+
+
 Square = int
 SQUARES = [
     A1, B1, C1, D1, E1, F1, G1, H1,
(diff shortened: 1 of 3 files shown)


niklasf:
Thanks, I think this is a good idea. Is `PromotionMoveError` in particular something that you want to distinguish? Otherwise I'd just merge that one with `IllegalMoveError`.

trevorbayless:
I was going back and forth on that myself. Personally, it is something that I'd like to distinguish but it does get difficult to catch on UCI moves, so it's probably best to just stick to the main three exceptions. The latest push removes `PromotionMoveError` and uses `IllegalMoveError` as it's replacement.

niklasf:
Alright. Let's go with that for now. Thanks!

PR #944, opened 2022-12-18
Make relative WDL consider model and ply

ianfab:
Got surprising behavior when changing the win-rate model on relative WDL. Looks like a bug to me.


niklasf:
Oops. Yeah, that's definitely a bug.

PR #954, opened 2023-01-26
Update cp multiplier to match Lichess implementation

jacksonthall22:
Last July, Lichess tweaked its formula for calculating "win chances" from centipawn score: https://github.com/lichess-org/lila/pull/11148


niklasf:
Thanks!

PR #957, opened 2023-02-03
Method for sending opponent information to engines

MarkZH:
This PR sends Xboard engines their opponent's name and rating after the `new` command and before the commands that set up the first move of the game. It would be used like this:

```python
engine = chess.engine.SimpleEngine.popen_xboard(...)
opponent_info = {"name": "BOT skakoui",
                 "rating": "1679 1356", # engine rating, then opponent rating
                 "computer": True}
engine.configure(opponent_info)

board = Board()
while not board.is_game_over():
    move = engine.play(board, Limit(time=10))
    board.push(move)
    board.push(get_move_from_internet())
```
This would result in the following communication just before the engine starts thinking about the first move:
```
<XBoardProtocol (pid=2948)>: Connection made
<XBoardProtocol (pid=2948)>: << xboard
<XBoardProtocol (pid=2948)>: << protover 2
<XBoardProtocol (pid=2948)>: >> feature usermove=1 sigint=0 reuse=1 myname="Genetic Chess #704622" name=1 ping=1 setboard=1 colors=0 done=1
<XBoardProtocol (pid=2948)>: << new
<XBoardProtocol (pid=2948)>: << name BOT skakoui
<XBoardProtocol (pid=2948)>: << rating 1679 1356
<XBoardProtocol (pid=2948)>: << computer
<XBoardProtocol (pid=2948)>: << force
<XBoardProtocol (pid=2948)>: << st 10.0
<XBoardProtocol (pid=2948)>: << post
<XBoardProtocol (pid=2948)>: << easy
<XBoardProtocol (pid=2948)>: << go

```
I modeled the changes on how the `computer` option was handled.

Previously, I would use `engine.protocol.send_line()` commands to manually inform Xboard engines of an opponent's name and rating. See [here](https://github.com/lichess-bot-devs/lichess-bot/blob/56cc1d028bd13fd4156913e47fc4856e9ffaae5a/engine_wrapper.py#L456) for an example from lichess-bot. However, information sent like this is sent before the `new` command, unlike every other chess interface. Without this modification, the only other time to send opponent information would be after the first move, which is not ideal since the opponent's rating might change how the engine plays.


niklasf:
Thanks, looks like a good way to provide this information.

I don't like that `rating` requires a special format, with two integers separated by a space. Could we split that into two parts?

MarkZH:
Now the ratings are separate entries ([code from lichess-bot](https://github.com/MarkZH/lichess-bot/blob/5e4b6dbadaa7fa9096cbeabbd0533f1d1f73ac62/engine_wrapper.py#L456)):
```python
engine.configure({"name": f"{game.opponent.title or ''} {game.opponent.name}".strip(),
                  "engine_rating": game.me.rating or 0,
                  "opponent_rating": game.opponent.rating or 0,
                  "computer": game.opponent.title == "BOT"})
```
Nobody's had any complaints about a similar problem with UCI engines. The current method for informing UCI engines of their opponents looks [like this](https://github.com/lichess-bot-devs/lichess-bot/blob/5b4c3192978b15223d966db55ad7bb4df243f531/engine_wrapper.py#L397):
```python
        name = game.opponent.name
        if (name and isinstance(self.engine.protocol, chess.engine.UciProtocol) and
                "UCI_Opponent" in self.engine.protocol.config):
            rating = game.opponent.rating or "none"
            title = game.opponent.title or "none"
            player_type = "computer" if title == "BOT" else "human"
            self.engine.configure({"UCI_Opponent": f"{title} {rating} {player_type} {name}"})
```
Do you want a similar change for UCI engines in python-chess as this PR? That is, so `UCI_Opponent` commands are sent after `ucinewgame`?

jdart1:
Regarding UCI: UCI options generally are "sticky:" they stay set until reset, and in that case the order of ucinewgame and the UCI_Opponent option should not matter. I have tried to see how existing interfaces handle this, but some including Winboard do not send "ucinewgame" at all, and as far as I can tell, Winboard also does not send UCI_Opponent in ics mode with a UCI engine.

niklasf:
Ah, so there's an UCI equivalent of sorts. Do you think it's worth trying to build an abstraction over both?

MarkZH:
Something like this with suitable overrides for UCI, XBoard, and `SimpleEngine`?
```python
class Protocol:
    @abc.abstractmethod
    async def send_opponent_information(self, *, opponent_name: str = None, opponent_rating: int = None, opponent_title: str = None, opponent_is_engine: bool = False, engine_rating: int = None) -> None:
        """
        Sends the engine information about its opponent. The information will
        be sent after a new game is announced and before the first move.

        :param opponent_name: Optional. The opponent's name.
        :param opponent_rating: Optional. The opponent's ELO rating.
        :param opponent_title: Optional. The opponent's title, such as GM, IM, and the like.
        :param opponent_is_engine: Optional. Whether or not the opponent is another engine.
        :param engine_rating: Optional. This engine's own rating. Only used by XBoard engines.
        """
```

jdart1:
That looks ok to me.

MarkZH:
If the configuration checks are kept in the python-chess library, then the `send_opponent_information()` method becomes protocol-agnostic. The user doesn't need to know the details of how an engine asks for information.

niklasf:
Looks good to me.

Can I ask for one more change? How about making an `Opponent` class to group these fields and have a stateful `engine.send_opponent_information(opponent: Opponent)` analogous to `engine.configure()`, but also a stateless `engine.play(..., opponent: Opponent)` analogous to `engine.play(..., options)`.

MarkZH:
I can do that. How should the `engine_rating` argument be handled? The Xboard `rating` command sends both the opponent's rating and the engine's rating (I can't think of why this is). The easiest way would be to include it in the `Opponent` class, even though it's not really information about the opponent.

niklasf:
Ah, that's slightly awkward. Would always sending `0` for that have any consequences?

MarkZH:
Depends on what each engine does with the parameter. How about keeping it as a separate parameter: `send_opponent_information(..., opponent: Optional[Opponent], engine_rating: Optional[int])` and similarly for `play()`?

MarkZH:
I've created the `Opponent` class and used it in the existing methods, keeping `engine_rating` as a separate argument. I need to think about how to incorporate it into `play()`. Two questions:

1. What do you mean by "stateless" when sending options to `play()`?
2. Should this option be added to `analysis()`, as well?

niklasf:
1. A second `play()` call is not influenced by state from a preceding `play()` call (except pondering and hash table in the same game). If the first call sets some options, the second call resets the options, unless they passed again.
2. Let's start with no.

MarkZH:
I think I've implemented the `play()` method parameter correctly.

I'm not sure I understand how the `options` parameter in this method is supposed to leave the engine unchanged after the move. Doesn't calling `self._configure(options)` within the `start()` methods of the `PlayCommands` change the `self.config` member of each `Protocol` class? There also doesn't seem to be any code to restore the settings in the engine that python-chess is controlling. I'm open to correction.

MarkZH:
I've added tests to document how I think the `opponent` argument to `play()` should behave. Changing an opponent mid-game results in a `ucinewgame` command for UCI engines and a `new` call for Xboard engines.

MarkZH:
Just to make sure I understand the configuration data members in the Protocol classes:
- `options` is the list of valid options for an engine and is populated during engine initialization and only read from afterwards to make sure only valid options are sent to the engine.
- `target_config` is the store of permanent configuration changes that persist from move to move.
- `config` is the temporary store of configurations that last only one move and is modified by passing a `dict` or similar to the `options` parameter of `play()` and `analyze()`.

If I have this right, then I just need to fix the `XBoardProtocol` to handle new opponents the same way as the latest commit. Then, this PR should be done.

MarkZH:
This PR is ready for review.

niklasf:
Looks great, thank you!

PR #962, opened 2023-02-15
strip option args

Disservin:
Strip leading and trailing whitespaces.
Currently implementation works just fine, however there are "buggy" engines which send another whitespace after the final token which is not required.

### DIFF
--- a/chess/engine.py
+++ b/chess/engine.py
@@ -1377,7 +1377,7 @@ class UciProtocol(Protocol):
                 current_var = None
                 var = []
 
-                for token in arg.split(" "):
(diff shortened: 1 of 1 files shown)


niklasf:
> arbitrary white space between tokens is allowed

-- https://backscattering.de/chess/uci/#whitespace

So i guess engines are actually allowed to do this, and it's a bug in this library, instead.

niklasf:
Merging, because the patch improves the situation, but it looks like engines would be allowed to use even crazier whitespace, which would still not be handled correctly.

PR #968, opened 2023-02-24
SVG output: add option for borders around board and diagram

jsonn:
This is useful for printing when the margin is typically white to still give the visual clues about separation. Option defaults to off.

Also slightly pad the coordinates at the top, so that "f" doesn't run into the outer border.


niklasf:
Thanks! Looks good with and without coordinates.

jsonn:
Would you be interested in a patch to avoid the overdrawing? I.e. using separate paths for outer border, margin and inner border? Advantage is that transparency settings for custom colors would work properly.

niklasf:
Ah, I didn't think about transparency. Would make sense to fix that, imo.

PR #969, opened 2023-02-25
SVG: use a stroked rect for borders and margin

jsonn:
This avoids painting the area under the squares. This makes it easier to use opacity for the field colors.

Strictly speaking, this is a BC change as transparent field colors will show the background of the parent element now and not the margin color, but that is more useful behavior to me.


niklasf:
Thanks! I wouldn't consider visual changes in `chess.svg` to be BC breaking, so all good.

Issue #722, opened 2021-01-01
Rust or Cython to speed up core components


jnewb1:
Has any thought or development gone into using cython for this library, especially the core components which could benefit from a speedup (move generation and validation)? I think that might be a good way to speed up the library and any chess engines that use this library. 

I tried just compiling just the core components using cythonize, but main issue right now appears to be lack of support for Generic and TypeVar from cython's end. (see [https://github.com/cython/cython/issues/2753](https://github.com/cython/cython/issues/2753))

I think if we get a basic cython build running and show the performance delta, that might be a good start (I'm interested in working on that, I was curious if anyone else had done something similar yet).

niklasf:
I previously rejected ideas diverging from pure Python to retain more flexibility, but now that we're on version 1.x it might be time to reconsider.

Other similar ideas were to use C/C++/Rust for some performance critical paths.

jnewberry0502:
Awesome! Maybe I'll try to get a cython build of the move generation working to see what kind of performance increases we could get. I do agree that pure python is much cleaner, but the speed improvements could be substantial, A small C/C++ library would also do nicely, or perhaps we could use an existing chess library and just add an option to use that library rather than the python core components. If anyone else is interested in this or has other ideas, please leave them here.

justinkterry:
Wouldn't C/C++/Rust make more sense than Cython?

jnewberry0502:
Well the idea with cython is that most of your python code remains the same, but yea I think a C/C++ helper library would also make sense

sudochop:
+1 If others are like me, this is a huge want as I use this library for processing large pgn files and my only solution for speedups is renting beefier vms.

niklasf:
Potentially useful: A C++ translation of the core module: https://github.com/pjpuzzler/cpp-chess (thanks for sharing @pjpuzzler!)

pawngrubber:
@niklasf I don't know enough about this repository to know where the bottlenecks are, but I know enough about Cython to take a swing at writing a PR.  Can you point me in the right direction for identifying bottlenecks?  For now I'm mostly interested in faster PGN reading and move generation (for example speeding up perft).  Also, if I want to write any tests for my PR where should I start?

niklasf:
@pawngrubber Maybe profiling `python examples/perft/perft.py examples/perft/random.perft` could show some points of interest. Other than that I don't really know what to do, except trail and error.

[...]

PR #984, opened 2023-03-30
Add `pgn.BaseVisitor.begin_parse_san()` method

snowdrop4:
Add a new `pgn.BaseVisitor` method: `begin_parse_san()`.

This is called when the parser encounters a SAN, but before the very expensive step of calling `board.parse_san()` on the SAN (which entails generating all the possible legal moves).

If the `begin_parse_san()` method returns `SKIP`, parsing of the SAN is skipped.

There are a couple of potential use-cases:

- only parsing the first $x moves from each game mainline (perhaps you're only interested in the opening stages of each game?)
- only parsing the first $x moves from each variation (perhaps the follow-up moves inside each variation do not matter to you?)
- trawling through lots of PGNs to find games where a specific opening sequence was played

As parsing the moves is by far the most time-consuming part of parsing a PGN, skipping any moves can result in huge time-savings.

On a real-world codebase that reads in PGNs with variations specified for most moves (and does a lot of computations afterwards), I'm seeing around a 20% speedup in *overall* program runtime, entirely from skipping parsing all but the first move of each variation.

### DIFF
--- a/chess/pgn.py
+++ b/chess/pgn.py
@@ -1043,6 +1043,13 @@ class BaseVisitor(abc.ABC, Generic[ResultT]):
         """Called after visiting game headers."""
         pass
 
+    def begin_parse_san(self, board: chess.Board, san: str) -> Optional[SkipType]:
+        """
+        When the visitor is used by a parser, this is called at the start of
+        each standard algebraic notation detailing a move.
+        """
+        pass
+
     def parse_san(self, board: chess.Board, san: str) -> chess.Move:
         """
         When the visitor is used by a parser, this is called to parse a move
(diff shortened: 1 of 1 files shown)


niklasf:
Thanks, this makes a lot of sense. In the more common case where every move is needed, how much does calling an extra method for every move cost?

snowdrop4:
Reading a PGN with 74 games in it, 100 times. Measured with `timeit`.

Master:

```
run 1 = 131.732180291001
run 2 = 132.3093085839937
run 3 = 131.67879545799224

mean = 131.906761444
```

PR branch:

```
run 1 = 132.53046183301194
run 2 = 132.58470095800294
run 3 = 132.302952500002

mean = 132.472705097
```

For an increase of 0.004% in total runtime.

Python 3.11.2 on macOS 13.3.1 on an M1 MacBook air.

snowdrop4:
Bonus benchmark, run the same way on the PR branch, but making use of the `begin_parse_san()` visitor to skip all but the first move from each variation:

```
run 1 = 95.46104741700401
run 2 = 95.54273075000674
run 3 = 95.95566983299796

mean = 95.6531493333 seconds
```

snowdrop4:
Semi-unrelatedly, but would you be amenable to a PR that ported the move generation to Rust with PyO3?

niklasf:
Thanks again.

Regarding #722, I am not sure, honestly. Priority would memory safety and correctness, of course, followed by simplicity of the build/release process. I've also considered ditching the aged python-chess API and making bindings for [shakmaty](https://crates.io/crates/shakmaty), instead.

PR #986, opened 2023-04-08
Add square_manhattan_distance and square_knight_distance functions

AInnuganti:
The chess library includes a function called square_distance, which calculates the Chebyshev distance (number of king steps) between 2 squares. This way of measuring distance is not always sufficient though, so I added two new distance functions.

- square_manhattan_distance calculates the Manhattan or Taxicab distance between 2 squares. This is the minimal number of orthogonal king moves it would take to travel from one square to another.

- square_knight_distance calculates the knight distance (using a pretty weird formula), which is the minimal number of knight moves to travel from one square to another

### DIFF
--- a/chess/__init__.py
+++ b/chess/__init__.py
@@ -220,25 +220,18 @@ def square_distance(a: Square, b: Square) -> int:
     """
     Gets the Chebyshev distance (i.e., the number of king steps) from square *a* to *b*.
     """
-    if a == b:
-        return 0;
     return max(abs(square_file(a) - square_file(b)), abs(square_rank(a) - square_rank(b)))
 
 def square_manhattan_distance(a: Square, b: Square) -> int:
     """
     Gets the Manhattan/Taxicab distance (i.e., the number of orthogonal king steps) from square *a* to *b*.
     """
-    if a == b:
-        return 0;
     return abs(square_file(a) - square_file(b)) + abs(square_rank(a) - square_rank(b))
 
 def square_knight_distance(a: Square, b: Square) -> int:
     """
     Gets the Knight distance (i.e., the number of knight moves) from square *a* to *b*.
     """
-    if a == b:
-        return 0;
-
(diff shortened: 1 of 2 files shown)


chess/__init__.py
```diff
@@ -218,10 +218,41 @@ def square_rank(square: Square) -> int:
...
+
+    dx = abs(square_file(a) - square_file(b))
+    dy = abs(square_rank(a) - square_rank(b))
+
+    if dx + dy == 1:
+        return 3
+    elif dx == dy == 2:
+        return 4
+    elif dx == dy == 1:
+        if 1 << a & BB_CORNERS or 1 << b & BB_CORNERS: # Special case only for corner squares
```
niklasf, line 250: the pattern used elsewhere for this is `BB_SQUARES[a] & BB_CORNERS`

chess/__init__.py
```diff
@@ -218,10 +218,41 @@ def square_rank(square: Square) -> int:
 
 def square_distance(a: Square, b: Square) -> int:
     """
-    Gets the distance (i.e., the number of king steps) from square *a* to *b*.
+    Gets the Chebyshev distance (i.e., the number of king steps) from square *a* to *b*.
     """
+    if a == b:
```
niklasf, line 223: let's not special-case `a == b` in any of the implementations

niklasf (changes requested):
Thanks! Looks like a useful and consistent addition.

Some tests in `test.py` (`class SquareTestCase`) would be good, especially for the knight distance.

AInnuganti:
Alright both of those changes have been made. I can add some test cases later

AInnuganti:
I've added tests for `square_manhattan_distance ` and `square_knight_distance`, and just for good measure, also added some for `square_distance`. Everything seems to be in working order

niklasf:
Looks good. Merging.

PR #998, opened 2023-04-26
Send game result to engines

MarkZH:
When a game is done, this method can be called to inform the engine of the final moves and the result of the game.

XBoard engines receive the final move of the game followed by a message like, for example, `result 1-0 {White mates}`. The second field is one of `1-0`, `0-1`, `1/2-1/2`, `*` to indicate white winning, black winning, draw, or an incomplete game. The message in curly braces is optional and meant for humans to read.

If this method is called on a UCI engine, nothing happens, as those engines do not expect any endgame messages.


chess/engine.py
```diff
@@ -1246,6 +1246,26 @@ async def analysis(self, board: chess.Board, limit: Optional[Limit] = None, *, m
         and stopping the analysis at any time.
         """
 
+    @abc.abstractmethod
+    async def send_game_result(self, board: chess.Board, winner: Optional[Color] = None, game_ending: Optional[str] = None, game_complete: bool = True) -> None:
+        """
+        Sends the engine the result of the game.
+
+        UCI engines only recieve the final position of the board. XBoard engines
```
niklasf, line 1254: Mhh ... this sounds like this method would send it, but if I understand correctly it doesn't.

chess/engine.py
```diff
@@ -2904,6 +2959,12 @@ def analysis(self, board: chess.Board, limit: Optional[Limit] = None, *, multipv
             future = asyncio.run_coroutine_threadsafe(coro, self.protocol.loop)
         return SimpleAnalysisResult(self, future.result())
 
+    def send_game_result(self, board: chess.Board, winner: Optional[Color] = None, game_ending: Optional[str] = None, game_complete: bool = True) -> None:
```
niklasf, line 2962: Let's make sure `game_ending` cannot be used to inject arbitary commands (`}\n...`).

Options [are safe](https://github.com/niklasf/python-chess/blob/58a54f135f0f1a6d9272e307fde7053f316a7b17/chess/engine.py#L288), but a similar thing slipped through wrt. `Opponent`.

MarkZH: Would getting rid of the newlines and curly braces be sufficient?
```python
game_ending = ' '.join(game_ending.split()).replace('{', '').replace('}', '')
```
Or, should an error be raised if the string contains new lines or curly braces?

MarkZH:
On another note, the Github tests are failing because the link to stockfish no longer works. Stockfish 11 can only be downloaded from a dropbox folder, which I doubt would allow automated downloads for long. The new link to Stockfish 16 is: https://github.com/official-stockfish/Stockfish/releases/download/sf_16/stockfish-windows-x86-64-avx2.zip

Github also has a cache mechanism to store files between test runs to negate the need to redownload or recreate them. We've done this in lichess-bot with [this commit](https://github.com/lichess-bot-devs/lichess-bot/commit/5808f5f311193c2658bbd42f22e6295c609e1bd6).

chess/engine.py
MarkZH: I went with raising an exception.

MarkZH: > Let's make sure `game_ending` cannot be used to inject arbitary commands (`}\n...`).
> 
> Options [are safe](https://github.com/niklasf/python-chess/blob/58a54f135f0f1a6d9272e307fde7053f316a7b17/chess/engine.py#L288), but a similar thing slipped through wrt. `Opponent`.

Does `send_opponent_information()` need fixing? The UCI version goes through `configure()`, so that should be safe. What about the XBoard version?

PR #1000, opened 2023-04-26
pgn ChildNodes: store boards under the hood, saving O(n) or O(n^2) op…

dubslow:
…s in some other methods

the movestack or plycount may also be of some use, tbd

### DIFF
--- a/chess/pgn.py
+++ b/chess/pgn.py
@@ -634,6 +634,7 @@ class ChildNode(GameNode):
 
     def __init__(self, parent: GameNode, move: chess.Move, *, comment: str = "", starting_comment: str = "", nags: Iterable[int] = []) -> None:
         super().__init__(comment=comment)
+        self._board = None
         self.parent = parent
         self.move = move
         self.parent.variations.append(self)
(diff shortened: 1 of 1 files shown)


niklasf:
We've come full circle (6330b5e8b6705d0b0f49f605707b1a7a4abeb688, 73644aca99c178bbbd6b9543b84667c66888d308, 007127635861b659e6d7692a85cae6260d803651). Caching like this will use quadratic memory instead, and due to the copy of each move stack, also still quadratic time. It is faster in practise, but also catastrophic for use-cases that keep a lot of games in memory.

There's probably not a great solution with the current data model.

I guess we should at least add a big warning to the documentation, maybe mentioning efficient workarounds:

* [`visit_board()`](https://python-chess.readthedocs.io/en/latest/pgn.html#chess.pgn.BaseVisitor.visit_board), which is less defensive about mutability and does not copy.
* Manually maintaining `board` like a cursor:
  ```python
  board = game.board()
  for move in game.mainline_moves():
      board.push(move)
  ```

PR #1003, opened 2023-04-30
Add optional identifier to Limit instances

MarkZH:
This new field in the Limit class is used to signal when the time control for a game has changed. XBoard engines expect to receive a `level` or `st` command to indicate the time control at the start of the game, and then only `time` and `otim` commands before each subsequent move to update the engine's internal clocks. The `level` and `st` commands should only be resent when the time control changes--like in FIDE tournament time controls. During a single time control, the same `clock_id` should be used for every Limit instance. This field has a similar function to the `game` parameter of the `play()` method.

Issue with previous discussion: #742

### DIFF
--- a/chess/engine.py
+++ b/chess/engine.py
@@ -332,6 +332,14 @@ class Limit:
     *white_clock* and *black_clock* are, then it is sudden death.
     """
 
+    clock_id: object = None
+    """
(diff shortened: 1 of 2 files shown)


niklasf:
Thanks. I like this solution.

PR #1013, opened 2023-06-05
A few minor cleanups

johndoknjas:
Just a few small cleanups I made going through the files. I also noticed in test.py, the `generate_legal_moves` function of `MockBoard` has a yield right after a return. I tried removing it, but then the checks gave me errors.


niklasf:
Thanks!

Any `yield` in a function makes it a generator, so `return; yield` is a trick to create an empty generator.

PR #1021, opened 2023-07-13
Error when XBoard opponent has newline in name

### LINKED ISSUE #1018: Review send_opponent_information() for UCI/XBoard command injection
https://github.com/niklasf/python-chess/pull/998#pullrequestreview-1520733825

Could actually be already safe, but best to double check.

MarkZH:
Use the configuration machinery to send the opponent's information to the engine instead of using the Opponent class directly. This will make sure that an exception is raised if the opponent's name has newlines in it before send_line() is called. This prevents opponents from injecting commands into the engine. For example, if an opponent sets their name as "Cheat\neasy", then the following commands will be sent to the local engine

name Cheat
easy

which would cause the engine to turn off pondering.

Also, add tests to show the change works as expected.

Fixes #1018

### DIFF
--- a/chess/engine.py
+++ b/chess/engine.py
@@ -2152,6 +2152,7 @@ class XBoardProtocol(Protocol):
 
     def _new(self, board: chess.Board, game: object, options: ConfigMapping, opponent: Optional[Opponent] = None) -> None:
         self._configure(options)
+        self._configure(self._opponent_configuration(opponent=opponent))
 
         # Set up starting position.
         root = board.root()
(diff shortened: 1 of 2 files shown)


MarkZH:
In [the comment for the `play()` method](https://github.com/niklasf/python-chess/blob/77bb7b2eaf66146b65cb8123b0d91eaf3de40d57/chess/engine.py#L1188), it says that if the `options` parameter is used to change the configuration of the engine, the original configuration would be restored after the move. Where does that happen? I want to make sure that the opponent information is restored, as well.

niklasf:
The comment is not entirely true, and should maybe be corrected/clarified: The engine wrapper behaves as if it were true, but the restore does not happen immediately after `play()`. Instead, on the next call to `play()` the current configuration will be compared to the target configuration again. If any option is not given again, then, it would be restored at that point.

PR #1032, opened 2023-08-02
Add PGN TimeControl Header Parser, closes #1031

### LINKED ISSUE #1031: Parse TimeControl PGN Header
It would be nice to have an option to parse the "TimeControl" PGN header and get the corresponding info (if applicable).
Something like 
```python
# e.g. for [TimeControl "40/7200:20/3600:1800"
> game.time_control()
[
    {"moves": 40, "time": 7200, "increment": 0, "delay": 0},
    {"moves": 20, "time": 3600, "increment": 0, "delay": 0},
    {"moves": 0, "time": 1800, "increment": 0, "delay": 0},
]
```
(Ideally the official [spec](http://www.saremba.de/chessgml/standards/pgn/pgn-complete.htm#c9.6) would be fully implemented).

The motivation behind this is that often the clock times noted in PGN files don't take the time bonus (or increment/delay) into consideration, so you might get something like `41... Ra1 { [%eval 0.04] [%clk 0:05:55] }` when the actual time left should be 5mins + time bonus.

mauricesvp:


niklasf:
Thanks!

PR #1036, opened 2023-08-12
Fixed generate_castling_moves to_mask bug

pjpuzzler:


niklasf:
Hi. I am guessing this is about Standard vs. Chess960 representation of castling moves? If so, this is by design.

PR #1043, opened 2023-09-14
Lenient whitespace parsing for engines

### LINKED ISSUE #970: UciProtocol is too strict about whitespace
https://backscattering.de/chess/uci/#whitespace gives engines a lot of freedom with regard to whitespace. Probably more than `chess.engine.UciProtocl` can currently handle.

I am not aware of any engine that makes use of this.

MarkZH:
Fixes #970

This PR is an attempt at reducing the number of assumptions about what whitespace can appear in messages received from engines. This is done in several parts:
1. A new function `_next_token()` extracts the next whitespace delimited word from a string, returning it with whitespace stripped out along with the rest of the line unchanged. This allows for string equality comparisons instead of `str.startswith()` and for the whitespace in the remaining string to be untouched, unlike with `str.split()`.
2. UCI options are parsed with a regex split that splits off parameter names without changing the whitespace in the parameter values.
3. If a received line is expected to exactly match another string, it will be compared with a `strip()` version of it: `if line.strip() == "readyok"`.

Other repeated code is refactored into functions.

### DIFF
--- a/chess/engine.py
+++ b/chess/engine.py
@@ -1324,7 +1324,7 @@ class UciProtocol(Protocol):
                 current_var = None
                 var = []
 
-                for token in arg.strip().split(" "):
+                for token in arg.split():
                     if token == "name" and not name:
                         current_parameter = "name"
                     elif token == "type" and not type:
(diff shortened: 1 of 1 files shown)


niklasf:
Perfect! Thanks.

PR #1052, opened 2023-10-17
replace '<object> == 0' by 'not <object>' & updated f-strings.

ProgramR4732:
Made the following changes:

-> replaced '<object> == 0' -> 'not object' statements for better runtime.
-> replaced a classic string by f-string for better readability & runtime.


niklasf:
Sorry, in code where the values are really semantically numbers I prefer to compare them as numbers.

PR #1062, opened 2024-01-23
feat: Adds support for displaying NAGs in the chess.svg.board AND other optimizations chess.svg.board function

Vikasg7:
* Adds support for NAGs in the svg.  
  - five NAGs has been added ie. !, !!, ?, ?? and ?!
  svg
* Attempted to remove unnecessary if statements from `chess.svg.board` and separated the generation of X squares from the rendering of pieces.


niklasf:
Nice addition, thanks!

Maybe a `nags: Iterable[int]` parameter would be more future-proof? We can currently only display one, but it could at least pick the most relevant one and show it. It would also be easier to just pass `node.nags` then, without having to filter out all the irrelevant ones (e.g., `Ke2 $70 $6 $77`).

Vikasg7:
Hi,
I just covered 5 out of 6 [here](https://github.com/Vikasg7/python-chess/blob/7e330b5003eea8c1e8e7aa35809a18d6cbbffd6e/chess/svg.py#L50) (6th being the speculative which I couldn't find the SVG for). Also, I am not sure how we are gonna decide the relevant ones in case of multiple nags. Also, I forgot to mention that we are covering only 5 in the function docs.   

So if we decide on displaying multiple nags, it would be difficult to decide where to show them in the square as I am already moving the nag left or right based on where the arrow is coming from so that it doesn't overlap the arrow.  

Glyphs show up at the end of the attached video which I am trying to produce with python.

https://github.com/niklasf/python-chess/assets/14152817/dcbb545e-2b71-4439-99d1-37d345437e33

S1M0N38:
Any update regarding NAGs rendering? Lichess rn supports NAGs in the studies:

Screenshot 2025-09-26 at 15: 43: 13

Is there something that I can do to merge this PR into the main?

PR #1068, opened 2024-02-23
Handle multiple comments on a move/variation/game

### LINKED ISSUE #946: Is there any way to write multiple comments on a move?
I'm trying to use python-chess to merge two pgn games, which is working fine. But it appears some PGN viewers really want the normal text comments and the %cal and %csl annotations separated into their own comment, like so:

 1. e4 { Here is a text comment } { [%cal Ge2e4] }

I seem to have no way to create this type of comment since the "comment" field in the GameNode class in pgn.py is a string, when I think I want it to be a list of strings. So I end up with this:

 1. e4 { Here is a text comment [%cal Ge2e4] }       <--- without a closing and opening bracket between the text and annotation

I have also tried setting node.comment = f"{text} }} {{ {annotations}" , ie injecting a } and a { inside the comment but it doesn't work, and is hacky anyway. It appears to insert the { but the } never appears.

What's your take on this? Just something the library doesn't support?

MarkZH:
Some chess programs that write PGN files (including lichess.org and others mentioned in the linked issue) will write multiple consecutive braced comments--e.g., `1. e4 { A very common opening } { Perhaps too common ... } { [%clk 0:00:01] }`. Previously, when read by `chess.pgn.read_game()`, these comments would be joined into a single comment with a single space between them. This PR creates a new class to read, store, and write these comments while keeping them separate. The user interface is kept backwards compatible by using `@property` decorators to allow assignment and comparison of bare strings and string lists to game node comments.

Closes #946

### DIFF
--- a/chess/pgn.py
+++ b/chess/pgn.py
@@ -197,13 +197,13 @@ class GameNode(abc.ABC):
     variations: List[ChildNode]
     """A list of child nodes."""
 
-    comments: list[str]
+    comment: list[str]
     """
     Comments that go behind the move leading to this node. Comments
     that occur before any moves are assigned to the root node.
     """
 
-    starting_comments: list[str]
+    starting_comment: list[str]
     nags: Set[int]
 
     def __init__(self, *, comment: Union[str, list[str]] = "") -> None:
--- a/chess/pgn.py
+++ b/chess/pgn.py
@@ -201,9 +201,8 @@ class GameNodeComment:
         """Create a string representation of the comments in PGN format."""
         comments = list(map(lambda s: s.replace("{", ""), self._comments))
         comments = list(map(lambda s: s.replace("}", "").strip(), comments))
-        output_comments = GameNodeComment(comments)
-        output_comments.remove_empty()
-        return "{ " + output_comments.join(" } { ") + " }"
+        comments = list(filter(None, comments))
+        return "{ " + " } { ".join(comments) + " }"
 
     def remove_empty(self) -> None:
         """Remove empty comments from the comment list."""
--- a/chess/pgn.py
+++ b/chess/pgn.py
@@ -184,6 +184,70 @@ class _AcceptFrame:
         self.in_variation = False
 
 
+class GameNodeComment:
+    """
+    PGN Comment Storage
+
+    A class that can hold one or more comments for a GameNode.
+    """
+    def __init__(self, comment: Union[str, list[str]] = ""):
+        self.set(comment)
+
+    def set(self, new_comment: Union[str, list[str]]):
+        """Replace the comment with a new comment or a list of comments."""
+        self._comments = new_comment if isinstance(new_comment, list) else [new_comment] if new_comment else []
+
+    def pgn_format(self) -> str:
+        """Create a string representation of the comments in PGN format."""
+        comments = list(map(lambda s: s.replace("{", ""), self._comments))
+        comments = list(map(lambda s: s.replace("}", "").strip(), comments))
+        output_comments = GameNodeComment(comments)
+        output_comments.remove_empty()
+        return "{ " + output_comments.join(" } { ") + " }"
+
+    def remove_empty(self) -> None:
+        """Remove empty comments from the comment list."""
+        self._comments = list(filter(None, self._comments))
+
+    def append(self, new_comment: str) -> None:
+        """Append a new comment to the end of the comment list."""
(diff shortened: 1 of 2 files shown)


MarkZH:
I tried to make `GameNodeComment` backwards compatible in terms of assignment and writing to a PGN file. The breaking change comes from getting comments from a node (with some allowances for comparing a single comment to a single string), since I didn't want users to accidentally miss comments. What would `GameNode.comment` return, just the first comment?

Saving this until 2.x would make things much easier. A new class wouldn't even be needed.
```python
class GameNode:
    comments: list[str]
```
If you want to save this PR until 2.x, this would be an easy change.

niklasf:
In 1.x, `GameNode.comment` would have to be the concatenation of all comments, to match the current behavior.

MarkZH:
That should do it. `GameNode.comments` returns the `GameNodeComment` (a `list[str]` essentially), and `GameNode.comment` returns a single `str` formed from joining the strings with spaces.

MarkZH:
Do you want to save this functionality for v2.0? I can make a new PR (not backwards compatible) that would be much simpler than this one--basically replacing all `comment: str` members of `GameNode` and `ChildNode` with `comments: list[str]` and changing the PGN comment functions.

niklasf:
Let's do 2.x. So it would contain "small" breaking changes and clean-ups, and more ambitious plans and far-reaching changes would be pushed back to 3.x.

MarkZH:
Updated changes:
1. Rename `GameNode` comment field from `comment` to `comments` and `starting_comment` to `starting_comments`.
2. Adapt methods that process comments (e.g., `eval()` and `set_eval()`).
3. Adapt PGN reading and writing to handle moves/games with multiple comments.

The change from `comment` to `comments` is the breaking change. Assigning to `node.comment` no longer has any effect. I figured that using a plural name emphasizes that the field should be a list. Plus, assigning a string comment instead of a list of strings will cause problems with comment processing methods (Item 2, above).

MarkZH:
Looks like the release of [mypy 1.11](https://mypy-lang.blogspot.com/2024/07/mypy-111-released.html) is catching new errors.

niklasf:
Fixed via #1100. Targetting [one final 1.x release](https://github.com/niklasf/python-chess/milestone/5) before making the jump to 2.x.

MarkZH:
Should extra whitespace be preserved when reading comments? To ask another way, what was the reason for [this commit](https://github.com/niklasf/python-chess/commit/f3367ea4b1402ed04be2190371ebc8c72c2276c8)? I would think that a newline in the middle of a curly brace comment was put there by word wrapping and so would not be a significant part of the comment. Is there a problem with formatting imported comments with the equivalent of
```python
comment = " ".join(comment.split())
```
to replace all whitespace with single spaces?

MarkZH:
I'm thinking of adding error messages for when users try to use the now-deleted `GameNode.comment` or `ChildNode.comment` instead of the new `GameNode.comments` or `ChildNode.comments`. Something like this:
```python
    _comment_error = AttributeError("GameNode.comment no longer exists. Use GameNode.comments to store a list of str comments.")

    @property
    def comment(self) -> None:
        raise GameNode._comment_error

    @comment.setter
    def comment(self, value: Any) -> None:
        raise GameNode._comment_error

    # And similarly for starting_comment
```
I'm thinking this could help catch errors in user code that would break when updating to v2.0.0. Then again, it would mask type checking errors if users tried to read or write to `node.comment` since that field would now exist. Thoughts?

niklasf:
It's an ugly hack, but maybe we can declare the methods only `if not typing.TYPE_CHECKING:`?

MarkZH:
I don't think one ugly hack can cancel out another ugly hack. I just tried type checking a simple example:
```python
class move:
    def __init__(self) -> None:
        self.comments: list[str] = []

m = move()
m.comments = "asdf"
m.comment = "dkfjdl"
```
Mypy came up with the following:
```
test.py:6: error: Incompatible types in assignment (expression has type "str", variable has type "list[str]")  [assignment]
test.py:7: error: "move" has no attribute "comment"; maybe "comments"?  [attr-defined]
Found 2 errors in 1 file (checked 1 source file)
```
So, users who type check with mypy should notice the change from `node.comment` to `node.comments`. Users who don't should see that changes they make to `node.comment` have no effect on the output. I think this means that the breaking change should be prominently reported in the v2.0 changelog--both the name change and type change.

PR #1084, opened 2024-05-17
Add rookognition

3d12:
Add [rookognition](https://github.com/3d12/rookognition) to the "selected projects" area of the README.

Thank you so much for maintaining this library! This was very easy to use, and very inspiring! I particularly love the Board.legal_moves generator, it makes it very easy to generate random legal board positions!


niklasf:
Thanks for sharing!

PR #1098, opened 2024-07-22
Update default time control from UNKNOW to UNKNOWN

deepyaman:
The `time_control` function docstring calls it `UNKNOWN`, and that makes more sense, so I assume this is just a typo.


niklasf:
Thanks!

PR #1104, opened 2024-09-05
SVG: Align pieces, check mark, selected squares and arrows when rendering a border

### LINKED ISSUE #1103: Align pieces to their squares center when generating SVG images with border
When generating board SVGs with a border around the board, then the border displaces the squares but not the pieces. That results in the pieces being drawn of center from their squares.

A lichess puzzle rendered with border, the pieces are of center to the top left:
3r3k_p5pp_2r2q2_2p4Q_8_8_P5PP_3R1R1K b - - 5 29_border

The same position without border, the pieces are centered:
3r3k_p5pp_2r2q2_2p4Q_8_8_P5PP_3R1R1K b - - 5 29_no-border

Lukasel:
When generating SVGs with a border some elements became missaligned. Now the same offset is used not just for the squares and coordinates, but also for the pieces, check mark, selected squares and arrows.

Resolves #1103

### DIFF
--- a/chess/svg.py
+++ b/chess/svg.py
@@ -283,6 +283,7 @@ def board(board: Optional[chess.BaseBoard] = None, *,
     inner_border = 1 if borders and coordinates else 0
     outer_border = 1 if borders else 0
     margin = 15 if coordinates else 0
+    board_offset = inner_border + margin + outer_border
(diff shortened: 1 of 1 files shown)


niklasf:
Thanks!

PR #1110, opened 2024-10-03
1109-pylint-interventions

evidencebp:
This PR fixes two minor pylint alerts.
It makes a long line shorther and removes unneccesary parentheses.

More detials appear in the commit mesage of each interventions.


niklasf:
Thanks! Cherry-picked the first. The one with the line-break makes the block slightly less readable, imo.

PR #1113, opened 2024-10-08
Fix error when parsing UCI options

### LINKED ISSUE #1112: Error while parsing UCI options
Since 1.11.0 there is an error when parsing UCI options. This occurs, for example, with Fairy-Stockfish:

```python
Exception parsing option min
Traceback (most recent call last):
  File "/home/torsten/.pyenv/versions/3.12.6/lib/python3.12/site-packages/chess/engine.py", line 1382, in parse_min_max_value
    return int(number) if number else None
           ^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: 'i'
```

The reason for this is that in the UCI option `UCI_Variant` the word `min` occurs in chess variants:

```
option name UCI_Variant type combo default chess var 3check [...] var mini var minishogi var minixiangqi [...] var torpedo
```

The following letter is an `i` and is therefore recognized by python-chess as the value for the `min` keyword.
Code to reproduce:

```python
import chess.engine

with chess.engine.SimpleEngine.popen_uci('fairy-stockfish') as engine:
    engine.configure({'Threads': 1})
```

MarkZH:
The regex that splits a UCI option into parts did not include word boundries, so words like "mini" would be split into "min" and "i", leading to parse errors. This commit puts word boundries into the regex to fix this.

Fixes #1112

### DIFF
--- a/.github/workflows/setup-ubuntu-latest.sh
+++ b/.github/workflows/setup-ubuntu-latest.sh
@@ -6,6 +6,9 @@ sudo apt-get install -y stockfish
 # Crafty
 sudo apt-get install -y crafty
 
+# Fairy-stockfish
+sudo apt-get install -y fairy-stockfish
+
 # Gaviota libgtb
 git clone https://github.com/michiguel/Gaviota-Tablebases.git --depth 1
 cd Gaviota-Tablebases
--- a/.github/workflows/setup-windows-latest.sh
+++ b/.github/workflows/setup-windows-latest.sh
@@ -1,6 +1,6 @@
 #!/bin/sh -e
 
-echo Download ...
+echo Download stockfish ...
 choco install wget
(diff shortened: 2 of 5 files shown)


niklasf:
Looks good. Let's also add a test for this.

MarkZH:
Not yet. I need to add fairy-stockfish to the test installation.

niklasf:
We could also use MockTransport with output from fairy-stockfish.

MarkZH:
Do you want to keep both tests or should I delete the fairy-stockfish test to reduce the external dependencies?

niklasf:
Let's keep it for now, since you already wrote the installation script and it looks like it should be pretty stable. Thank you, both!

Torom:
@niklasf Are you planning a 1.11.1 release with this fix or will we have to wait until 2.0?

niklasf:
I suppose 2.0.0 may be quite a while out, so published 1.11.1.

PR #1117, opened 2024-10-26
Add 'movesleft' parameter to UCI info parsing

mooskagh:
Add `movesleft` to the list of recognized `info` values.

`movesleft` is a prediction of the remaining game duration in ply, it's is output by Leela Chess Zero (since 3 years ago already), and always had to patch the python-chess in order to get it (there's no way to get "unknown" fields even as a string).

I guess it makes sense to upstream it instead.


niklasf:
Thanks!

PR #1118, opened 2024-10-27
added hash

eyalk11:
Added hash to Board class

### DIFF
--- a/chess/__init__.py
+++ b/chess/__init__.py
@@ -1584,6 +1584,8 @@ class _BoardState:
         board.halfmove_clock = self.halfmove_clock
         board.fullmove_number = self.fullmove_number
 
+#from  numba.experimental import jitclass 
+#@jitclass 
 class Board(BaseBoard):
     """
     A :class:`~chess.BaseBoard`, additional information representing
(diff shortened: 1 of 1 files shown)


chess/__init__.py
```diff
@@ -3817,14 +3819,21 @@ def _repr_svg_(self) -> str:
 
     def __eq__(self, board: object) -> bool:
         if isinstance(board, Board):
-            return (
-                self.halfmove_clock == board.halfmove_clock and
-                self.fullmove_number == board.fullmove_number and
-                type(self).uci_variant == type(board).uci_variant and
-                self._transposition_key() == board._transposition_key())
+            return hash(self) == hash(board)
```
niklasf, line 3822: `hash(self) == hash(board)` does *not* imply `self` equals `board` (there can be hash collisions).

niklasf:
From https://docs.python.org/3/reference/datamodel.html#object.__hash__:

> If a class defines mutable objects and implements an [`__eq__()`](https://docs.python.org/3/reference/datamodel.html#object.__eq__) method, it should not implement [`__hash__()`](https://docs.python.org/3/reference/datamodel.html#object.__hash__), since the implementation of [hashable](https://docs.python.org/3/glossary.html#term-hashable) collections requires that a key’s hash value is immutable (if the object’s hash value changes, it will be in the wrong hash bucket).

chess/__init__.py
eyalk11: I am not sure I understand the claim. Are you claiming the hash is weak? Because hash collisions are negligible.

niklasf: Maybe for your use case.

But this is a general purpose library. So we shouldn't assume that we will only ever see 2^32 positions (at which point collisions are likely even for strong 64 bit hashes), or even that inputs aren't specifically chosen by an adversary.

PR #1126, opened 2024-12-24
Split Tests into Separate Files

axolotl-logic:
I sat down to write some tests and was alarmed at the test file being thousands of lines long. So I split it up. Is this something you'd want? Tests still run locally.

I didn't commit it, but I wrote this script for generating test coverage.

```bash
#!/usr/bin/env sh

# Usage: scripts/generate-coverage-report.sh

coverage run -m unittest discover
coverage html
coverage report
echo "## Open htmlcov/index.html for detailed breakdown ##"
```


niklasf:
Sorry, I like the combined file.

axolotl-logic:
No problem!

PR #1130, opened 2025-01-05
gaviota.py: simplify and improve code

adbar:
Hi @niklasf, I was having a look at the code to understand how endgame databases work and I saw that the Gaviota part of the code could be improved. Here is a first step:

- Duplicate lines removed
- Reduced code complexity (good for maintenance...)
- Convenience functions introduced to avoid repetitions

The whole structure is the same so the PR should hopefully be easy enough to review. The tests pass and the code should also run faster.

### DIFF
--- a/chess/gaviota.py
+++ b/chess/gaviota.py
@@ -11,10 +11,11 @@ import os.path
 import struct
 import typing
 
-import chess
-
+from itertools import accumulate
 from types import TracebackType
 from typing import BinaryIO, Callable, Dict, List, Optional, Tuple, Type, Union
+
+import chess
 
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


adbar:
There are apparently lines not covered by the tests that mypy was able to catch. I am open to modifying the code if it needs improvement.

chess/gaviota.py
```diff
@@ -179,26 +166,15 @@ def init_ppp48_idx() -> Tuple[List[List[List[int]]], List[int], List[int], List[
...
-                    ppp48_idx[i][k][j] = idx
-                    ppp48_idx[j][i][k] = idx
-                    ppp48_idx[j][k][i] = idx
-                    ppp48_idx[k][i][j] = idx
-                    ppp48_idx[k][j][i] = idx
-                    ppp48_sq_x[idx] = i
-                    ppp48_sq_y[idx] = j
-                    ppp48_sq_z[idx] = k
+                    for p in [(i, j, k), (i, k, j), (j, i, k), (j, k, i), (k, i, j), (k, j, i)]:
+                        ppp48_idx[p[0]][p[1]][p[2]] = idx
```
MarkZH, line 176: You can use [`itertools.permutations()`](https://docs.python.org/3/library/itertools.html#itertools.permutations) to simplify this loop.
```python
for ii, jj, kk in itertools.permutations((i, j, k)):
    ppp48_idx[ii][jj][kk] = idx
```

adbar: That's correct, thanks for the idea.

niklasf:
Thanks, looks good! Will test offline with a larger test suite just to be sure.

niklasf:
Found #1132 (happy if someone wants to take a look), but that's not a regression. So merging this.

niklasf:
There appears to be a suble way this is not equivalent, but I don't have the bandwidth to track it down at the moment. Reverted for now via 636e95fbf292f322fc4ab31b8c4add51f7534362 (see commit message for the problem).

PR #1133, opened 2025-01-12
refactor __init__.py and add constants

adbar:
Hi @niklasf, as in my last PR I'm working my way through the code and trying to make it more readable and efficient. This one simplifies the init file and add constants to speed up things a bit.


niklasf:
Sorry, I don't like splitting the file. Also measuring a ~10% performance regression in `examples/perft/perft.py random.perft` instead of improvement.

PR #1179, opened 2026-02-03
Multiple kings per color (was_into_check and more affected)

### LINKED ISSUE #1169: `was_into_check` behavior on multiple kings
[`was_into_check` (L1995-L1997)](https://github.com/niklasf/python-chess/blob/master/chess/__init__.py#L1995-L1997)
```
    def was_into_check(self) -> bool:
        king = self.king(not self.turn)
        return king is not None and self.is_attacked_by(self.turn, king)
```

Here it only picks one king.... but if the position loaded has multiple kings, it only return the bool of one king which depends on FEN parsing order, that means - a flipped position can bring different king of the corresponding side to move.

While it isn't too much - this is kind of UB for pseudo-legal FENs, and it affects `Board.is_valid`, `Board.status` and possibly its variants.

winapiadmin:
Not only of https://github.com/niklasf/python-chess/issues/1169 that `Board.king()` function is completely wrong for multiple kings per color but `msb(king)` or `lsb(king)` don't make sense if there's multiple kings for one color.

### DIFF
--- a/chess/__init__.py
+++ b/chess/__init__.py
@@ -916,7 +916,7 @@ class BaseBoard:
         considered.
         """
         king_mask = self.occupied_co[color] & self.kings & ~self.promoted
-        return msb(king_mask) if king_mask and popcount(self.kings) == 1 else None
+        return msb(king_mask) if king_mask and popcount(king_mask) == 1 else None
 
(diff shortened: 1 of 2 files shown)


niklasf:
Thanks!

PR #1190, opened 2026-05-28
variation_san, include a space after the ellipsis for black moves

kraktus:
Feel easier to read and also how Lichess write its PGN


niklasf:
Thanks! Also now consistent with `chess.pgn`.

PR #1205, opened 2026-08-19
Fix `parse_rank()`, `rank_name()`

jacksonthall22:
I apparently committed this bug 11 months ago in https://github.com/niklasf/python-chess/pull/1163. Codex [found it](https://github.com/jacksonthall22/chess.ts/pull/24) while helping me build `chess.ts`, a transpilation of `python-chess` to typescript. I basically instruct it to go commit by commit and reproduce the diffs as faithfully as possible, notwithstanding syntax and a few language-specific patterns for which I document explicit instructions/conventions for how to transpile - but it's not supposed to make any decisions or changes, even if the python is wrong, unoptimized, etc. It initially, correctly, coded the bug into the typescript and left me a TODO to fix when it's patched in `python-chess`. Funny that otherwise I would have never noticed this 😛

### DIFF
--- a/chess/__init__.py
+++ b/chess/__init__.py
@@ -288,18 +288,18 @@ def file_name(file: File) -> str:
     """Gets the name of the file, like ``a``."""
     return FILE_NAMES[file]
 
-def parse_rank(name: str) -> File:
+def parse_rank(name: str) -> Rank:
     """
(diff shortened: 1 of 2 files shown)


niklasf:
Oops. Thanks!

